PR #22, opened 2015-01-13
Start work on destroyables, allow TNT touches on cores

iEli2tyree011:


twizmwazin:
@rywarsaw feel free to clean it up

iEli2tyree011:
@rywarsaw only because I want it easier to implement channels easier later and it's easier to understand than several and operators stuffed into one statement.

PR #23, opened 2015-01-13
Add universal ConsoleSender check, VersionUtil(Untested) and simplify permissions

rywarsaw:
This pull request adds a ConsoleSender check in Cardinal.java to prevent the console from running CardinalPGM commands. Currently I didn't  see any ConsoleSender checks, and this is easily implemented without too much logic. 

It also simplifies permissions, in a way that makes sense, adding permissions to commands that didn't have them. Great for public scrimmage matches where you don't really want random people to be joining the match. 

The VersionUtil is highly experimental, I don't know if it will work, I plan on testing it later today.


twizmwazin:
The VersionUtil class appears to be missing

rywarsaw:
@twizmwazin It should be back in there now. For some reason it didn't automatically add itself to the change list. Thanks!

twizmwazin:
Does the version checker return the version in the plugin.yml or the git revision?

rywarsaw:
@twizmwazin I believe it actually returns from the maven pom. I haven't tested it at all, so I don't know if it is working. If you want, I can test it out later today and see how it works.

twizmwazin:
I'll test it tonight and if everything is good I'll merge it then.

rywarsaw:
@twizmwazin 

```
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <mainClass>in.twizmwaz.cardinal.Cardinal</mainClass>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                        </manifest>
+                    </archive>
+                </configuration>
+            </plugin>
```

Is what I added into the pom, in order to theoretically make this work.

twizmwazin:
I think the universal console checker may not be a good idea, there are a number of non-player specific commands that could be used in the console. Only commands such as join should not be able to be used in the console. Also, I'm not too familiar with Bukkit's permission system, is there a way that that players could have permissions by default without having to change any configuration files? I don't want to require the end-user to need to edit permissions.

rywarsaw:
@twizmwazin Afaik both of those can be done. I would just make small built-in rank system that users can use for easy permissions.

- /add user <username> Adds a user with basic permissions to use the plugin.
- /add captain <username> Adds a captain with higher permissions such as /ready (if we do that)
- /add admin <username> Grants all permissions for cardinal (Assuming console access)

rywarsaw:
Closed while I do some changes, fix universal console checker, and try to implement a light weight permissions system.

twizmwazin:
I don't think that /add user should be required, by default users should be able to do basic functions such as join or view map information.

rywarsaw:
@twizmwazin I disagree, but its your plugin so do whatever you want.

twizmwazin:
My reasoning is that usually, scrims begin relitively disorganized, and if a team of 10 players is joining a server they've never played on, it would add unnessesary time to starting the scrim.

rywarsaw:
@twizmwazin I was mainly reflecting on the challenges that most scrimmage servers have, which is people who aren't on a team randomly joining. When You have teams of 16 players or more, its really challenging to keep track of who is on who's team. 

There needs to be a way to both make it extremely easy for the end user to use, but still allow it to be secure enough for your average team. I'm merely suggesting ways of doing so.

twizmwazin:
@rywarsaw A scrimmage wouldn't consist of more than teams of 10 usually. While enabling whitelist is always an option, EllyPGM has a command to enable and disable team switching, essentionally locking players to a team. This could be a good solution.

PR #46, opened 2015-01-18
Fix wool touches

CaptainElliott:
Previously there were two `onWoolPickup` events It also only would register when you picked it up from an inventory, now it registers the pick up even if you picked it up from the ground. I also broadcasted the wool touch.

the wool touches should be broadcasted to only players on the team and right now I don't see where you are tracking who is on what team, so if you are let me know and ill start work on team chat


CaptainElliott:
https://github.com/twizmwazin/CardinalPGM/pull/47

PR #47, opened 2015-01-18
Add Chat channels and Fix wool touches 

CaptainElliott:
I know some parts are a bit messy but it all works


twizmwazin:
Thanks!

iEli2tyree011:
What was the problem with the wool touches? They both worked for me before, it just looks like you switched the two around

CaptainElliott:
They would not work if you picked it off the ground, there was also two identical `onWoolPickup` methods

iEli2tyree011:
Why wouldn't they work? I'm not sure what the point of switching them around was, but it doesn't affect anything I suppose

CaptainElliott:
Nothing was switched around, there was two events of `InventoryClickEvent`.

`InventoryClickEvent` does not register when a player picks up the wool from item form so I added `PlayerPickupItemEvent` the diff looks weird though

PR #51, opened 2015-01-18
Changes to improve consistency and coding conventions

iEli2tyree011:


iEli2tyree011:
We're going to use a system with permissions to manage toggles. Also, you had it so that if I did /a, everyone would talk in admin chat, /g for everyone talking in global. We're saving all the player-specific settings until the implementation of permission management at a future date.

twizmwazin:
There will be a localization system soon

CaptainElliott:
It was set up so team chat was default so whenever you talked normally you would talk in team chat. If you did `/a msg` then it would end an admin message, but speaking normally would still be in team chat, unless you switched it with just `/a`

iEli2tyree011:
Yes, but when I did /a, it would set *everyone's* default channel to admin chat.

CaptainElliott:
ohhh yeah, I see

PR #104, opened 2015-01-24
Add Rage Module and start work on blitz

CaptainElliott:
Rage works fine.

Started work on blitz - the fundamentals and logic works, the messages need to be tided up and the scoreboard needs to be implemented as well as the win condition.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
@@ -28,6 +28,7 @@ import in.twizmwaz.cardinal.module.modules.monumentModes.MonumentModesBuilder;
 import in.twizmwaz.cardinal.module.modules.motd.MOTDBuilder;
 import in.twizmwaz.cardinal.module.modules.observers.ObserverModuleBuilder;
 import in.twizmwaz.cardinal.module.modules.projectiles.ProjectilesBuilder;
+import in.twizmwaz.cardinal.module.modules.rage.RageBuilder;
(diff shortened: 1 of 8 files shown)


iEli2tyree011:
We'd planned to make the life counter using LazyMetadataValues, do you think you could switch it to that?

iEli2tyree011:
If you need a reference, the KillStreak module uses them.

CaptainElliott:
Working on it as well as completing the blitz module

twizmwazin:
The map system works fine for blitz, it wasn't how I would have done it but if it works fine I don't see a reason why it is "wrong."

CaptainElliott:
How is it redundant? @twizmwazin

CaptainElliott:
I have already re-written it with the LazyyMetaData, I was unaware of it before, I prefer it to the hasmap

iEli2tyree011:
getChild returns any elements that have the name specified. Just check if != null.

CaptainElliott:
I did that before but It would not find the rage tag

iEli2tyree011:
Or, sorry, getChildren does that. getChild just returns the first element I believe.

iEli2tyree011:
How about get children("rage").size() > 0?

twizmwazin:
@CaptainElliott The redundant line was getting an element (or null) named "rage," and then if it was named rage do

@iEli2tyree011 Either getChildren("rage") in a for, or if getChildren("rage").size > 0 would be possibly safer, duplicate modules here shouldn't be an issue though.

CaptainElliott:
Ready to merge the TODOs still need to be done

PR #121, opened 2015-02-01
Add Spanish lang file

felipufeliu:
I may tomorrow add translations


iEli2tyree011:
We can just add it once we do the actual translations

PR #135, opened 2015-02-03
Add Proximity Alarms

CaptainElliott:
Still have to add the fireworks, but I plan on adding them when I add fireworks to when an objective is completed


twizmwazin:
How thoroughly was this tested?

CaptainElliott:
I tested the notify and detect functions with another player, they seemed to work

twizmwazin:
Awesome, thanks.

PR #136, opened 2015-02-03
Allow warning message to accept text formatting codes

CaptainElliott:
If you did not know warning messages can be colored 
https://github.com/OvercastNetwork/Maps/blob/3d91ac4019c8f07c7ff5486abc63923864f4f1e3/DTM/Standard/DATAbase/map.xml#L100


iEli2tyree011:
The color codes should be dealt with when the module is instantiated, not when it's being changed to a warning message. While this is a good safeguard, it should be already done before. Thank you for the change though!

PR #144, opened 2015-02-04
Add Block Explosion Physics

CaptainElliott:


twizmwazin:
This might need some work before I merge this. On my first test, I got [this] (http://hastebin.com/eqagiyiciq.avrasm) exception, and did not see any blocks. Also, is there a reason that you have this in the BlockDrops module rather than creating a separate module?

CaptainElliott:
It's part of the block drops module, I did not get any errors when I tested so I'll look into it.

CaptainElliott:
Out of Curiosity could I see how you were defining the region

twizmwazin:
I tested it on cos(tnt) right outside the west yellow spawn.

CaptainElliott:
Ah I see, cos does not define a region , so it's returning null. I also need to add filter support

CaptainElliott:
So I have fixed the errors, but the land chance is not working as expected and I can't think of another working way to do it

twizmwazin:
If land chance is the chance of the falling block to become a block when it hits the ground, you could store meta data and use that. I'm in school so I can't look at it.

CaptainElliott:
Not sure what just happened to this PR, ill fix it and create a new one.

CaptainElliott:
closing in favour of https://github.com/twizmwazin/CardinalPGM/pull/149

PR #154, opened 2015-02-08
TNT Module

CaptainElliott:
The instant ignite is doing some weird things (the power set does not seem to be right)
and as of now I have not though of a efficient way to do the yield


iEli2tyree011:
Can you look into fixing that before we merge this?

iEli2tyree011:
Also, you left quite a few debug messages in the TNT module.

PR #161, opened 2015-02-09
Updated /join error message after a match has ended

felipufeliu:


felipufeliu:
I fixed it , thanks @dentmaged 😄

PR #166, opened 2015-02-11
Change the order of cycling messages.

G61:


src/main/java/in/twizmwaz/cardinal/module/modules/match/MatchModule.java
```diff
@@ -55,9 +55,9 @@ public void onMatchEnd(MatchEndEvent event) {
 
     @EventHandler(priority = EventPriority.LOWEST)
     public void onCycleComplete(CycleCompleteEvent event) {
+    Bukkit.broadcastMessage(ChatColor.DARK_AQUA + "Cycled to " + ChatColor.AQUA + event.getMatch().getMapInfo().getName());
```
twizmwazin: The indentation is incorrect

twizmwazin:
What exactly is the benefit of this pr?

G61:
Just the "You joined observers" is appearing before Cycled to x and changing the order.

twizmwazin:
Sure, why not

PR #221, opened 2015-02-20
Fix for #217

### LINKED ISSUE #217: /g with no arguments sends blank global message
When a user runs /g with no arguments, it sends a blank message. I know that you have to work on toggling but I think it would be good to fix it.

/g

Version: Version

ParaPenguin:
This should fix #217, and @twizmwazin what were you thinking, not using `getJoinedStrings(0)`


twizmwazin:
@ParaPenguin Whoops, missed that method.

PR #226, opened 2015-02-21
Cleanup some item code

ShinyDialga45:
A lot of it was repeating, so I just centralized some code (I was tired doing this so point out anything I missed if there's any).


twizmwazin:
Awesome, thanks

PR #231, opened 2015-02-21
Tutorial

CaptainElliott:
Making this PR to see if anyone can tell me why the module collection for Tutorial only has an index of 0.

With that in mind, this currently does not work


dentmaged:
@CaptainElliott You need to add the Builder to the ModuleFactory. That should work ;)

dentmaged:
@CaptainElliott I just noticed how old the module builder is, try making a patch of the commits and apply that patch on a new version and edit it again to check for bugs.

CaptainElliott:
I am aware that this is old, I intend to redo both of the PRs I have open.

I will properly close this as I don't want it to be merged

PR #257, opened 2015-02-24
[Missing Feature] Add objetive map and name when cycle

AngelCasillas01:
http://imgur.com/u4Qieea 

I was on my iPad when screenshot taken, but i think you understand me


twizmwazin:
Why do people keep making pull requests from this branch to suggest features.

PR #285, opened 2015-02-26
Call block drops on EntityExplode, fixes #276

### LINKED ISSUE #276: Blockdrops don´t droping items when the block is destroyed by TNT
Video proof: http://youtu.be/aKRn3n5aYzE
Cardinal version: http://gyazo.com/bf4fe51e43e76c00bc9ce2d668b579db

CaptainElliott:
Sorry @iEli2tyree011 if you started it ;p has some free time so I fixed this :)


iEli2tyree011:
Thanks, didn't notice this before I pushed. TntTracker can sometimes return null when you did that, make sure you watch out for that as well.

PR #294, opened 2015-02-27
Add /recycle {time}

javipepe:
It should be working, can't check it right now but everything seems alright. It just sn's the current map when it's about to cycle.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/command/CycleCommand.java
+++ b/src/main/java/in/twizmwaz/cardinal/command/CycleCommand.java
@@ -41,6 +41,35 @@ public class CycleCommand {
             GameHandler.getGameHandler().startCycleTimer(30);
         }
     }
+    
+    @Command(aliases = {"recycle"}, desc = "Recycles to the same map", usage = "[time]", flags = "f")
+    @CommandPermissions("cardinal.match.cycle")
+    public static void cycle(final CommandContext cmd, CommandSender sender) throws CommandException {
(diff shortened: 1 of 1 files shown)


twizmwazin:
I don't like the idea of reusing the setnext and cycle commands. I'd rather you got the current loaded map and set that as the next map without dispatching the command. Then run the cycle timer.

CaptainElliott:
Its probably inefficient to call a command, best cycle it manually. Also you should there are many NPE's that could occur, such as a player not specifying a team, moreover the /recycle command does not accept any parameters and ends the game with whatever team is currently winning

javipepe:
Oh, I'll play with the <code>cycle()</code> method later and will push it on here

twizmwazin:
It is also worth noting that the build failed, please use a proper IDE such as IntelliJ or Eclipse, and push code using git, not GitHub's editor.

Also, you wouldn't want to `cycle()`, you would want to start the cycle timer.

dentmaged:
@javipepe Also replace line 47 with recycle (the void name), as cycle is used before in the class.

AngelCasillas01:
I think that if you unbroke your code, this can be merged :o

twizmwazin:
@AngelCasillas01 I'm not merging anything that does not build.

AngelCasillas01:
@twizmwazin on your recycle commit what have you used?

twizmwazin:
@AngelCasillas01 I wrote the command myself, this was not merged.

AngelCasillas01:
@twizmwazin Nice job i have tested already!

PR #315, opened 2015-03-01
Prevent /kill spawning you in match, fixes #308

### LINKED ISSUE #308: When you join into a team and you use /kill, you spawn with kit
When you join into a team and you use /kill, you spawn with kit, although the match not starting.

CaptainElliott:


iEli2tyree011:
Looks good

PR #345, opened 2015-03-06
Prevent GUIs to be displayed on max height, closes #342.

### LINKED ISSUE #342: Sign GUI is still displayed when placing signs at max build height
When placing a sign, like any other block, at the max build height for a map, the placement is cancelled. However, with a sign, the sign editor GUI is still displayed.

This can give a player the false impression that the sign was actually placed, leading to them entering text which will be discarded.

Steps to reproduce

1. Place any block at the maximum build height for a map.

2. Attempt to place a sign on top of said block.

3. Observe that the sign edit GUI is displayed.

iJorGeZz:
Small fix until SportBukkit is ready to solve completely this problem.


twizmwazin:
Thanks!

PR #356, opened 2015-03-08
Make sure the match hasn't started on /(un)ready

javipepe:
This will just make sure the match is not running when /ready or /unready commands are executed.


CaptainElliott:
Your build failed, also don't check it it is not playing, check if the game state is in `WAITING`, we don't want to be able to /ready when the match has ended or cycling

CaptainElliott:
https://github.com/twizmwazin/CardinalPGM/commit/1840ed24e5e00b22d48a71a48a1ac1063d26fec9

PR #357, opened 2015-03-08
Add automatic cycling, began work on module-based cycling and remove...

dentmaged:
...useless 'this.add'

Automatic cycling is disabled by default (as seen in the config: ```cycle: -1```)


CaptainElliott:
I don't think we need automatic cycling, this is just a scrim plugin, you need to remember that

CaptainElliott:
Also, there is no point removing the `this.` as it makes no difference what so ever, however it does make more sense and easier to read. The only reason to explicitly not to have it if the method `add` was static

dentmaged:
It makes the project smaller in size, it's unneeded and I find automatic cycling useful, but if you want to close it, feel free to. I'll create a new PR with only module-based cycling.

CaptainElliott:
@dentmaged using `this.` has no impact on size or performance

elespawol:
On tm.oc.tc there is not auto cycling... so i think its not necesary
El dia 08/03/2015 15.02, "Elliott" <notifications@github.com> va escriure:

twizmwazin:
I'm ok with the module based cycling, although I'm going to read over it a\carefully before I merge it. As Elliott said, I really care about code style and conventions, including using `this` whenever possible.

dentmaged:
Okay, should I make a new PR? :\

twizmwazin:
@dentmaged You can jeep this one, although don't have more than 2 commits. You can amend commits in git.

dentmaged:
@twizmwazin You can? How?

twizmwazin:
Merged

PR #364, opened 2015-03-10
Updated min of permanent bans 2>10

RubeGamerV:


iEli2tyree011:
There's no reason to add this, it breaks more than it fixes (even though there's nothing to be fixed). All players need to do is provide a player and a reason. If they want the reason to be one word, they can.

PR #392, opened 2015-03-13
Fix default chat channels' message.

javipepe:
All these photos were taken on overcast network servers:

Global:

Adminchat:

Team:


CaptainElliott:
I knew what OCNs said, we do not have to mimic all of there messages word for word, besides, I prefer what we already have as it is clearer to what channel you have selected (due to the channel being colored red) 

but I'll leave this up to @twizmwazin

javipepe:
Uh I see @CaptainElliott. I don't really see a huge difference in terms of clarity; I mean the color thing is a little bit pointless. 

Also there was a spelling mistake that I fixed on this commit as well (so if you want to keep the message as it was correct it :stuck_out_tongue: ) on the previous version (<code>defult</code> should be <code>default</code>)

CaptainElliott:
oh whoops :p

twizmwazin:
These messages really should be localized rather than hard coded in.

PR #395, opened 2015-03-14
Changed "Admin channel" to "Staff Channel"

RubeGamerV:


twizmwazin:
It isn't the staff channel, though, it is the admin channel.

PR #402, opened 2015-03-16
Added CardinalPGM in plugin.yml 

AngelCasillas01:
So by typing /plugins, appear CardinalPGM which has more logic


twizmwazin:
@AngelCasillas01 If you want an explination on why I closed this, it is because when the project is built with maven, all variables in the plugin.yml are replaced. So when the jar is compiled with `mvn clean install`, the plugin.yml is correctly filled in.

PR #406, opened 2015-03-17
Improve how invalid SportBukkit versions are handled

ryanwarsaw:
This is a quick fix for how invalid SportBukkit versions are handled. Instead of sending a bunch of errors, it should instead send them a message that they are using a invalid version of SportBukkit, and give them information on what the correct version would be. It then disables the plugin to prevent any further issues. This hasn't been tested yet, but is fairly simple so its presumed to be working.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/Cardinal.java
+++ b/src/main/java/in/twizmwaz/cardinal/Cardinal.java
@@ -15,7 +15,6 @@ import org.bukkit.Bukkit;
 import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.ConsoleCommandSender;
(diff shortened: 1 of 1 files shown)


CaptainElliott:
You might want to do `contains.1_8` because there are other 1.8 versions that will be released

ryanwarsaw:
@CaptainElliott I didn't realize I did that, I will fix it later tonight.

twizmwazin:
If this is going to be merged, I want to have a sportbukkit check as well, to prevent users from attempting to load it on craftbukkit & spigot.

ryanwarsaw:
@twizmwazin This is just a quick fix, although I have no doubts you could easily code that if you wanted too :)

ryanwarsaw:
Closed, as this feature is presumed to be added when a SportBukkit check is implemented.

PR #418, opened 2015-03-22
Fix wool touch message & remove permissions in proximity command

Alfalik:


iEli2tyree011:
Can you revert lines 23 and 44? I'd rather have a localized error message, along with permissions for being consistent with other commands and their permissions. I'll be happy to merge after that, though.

Alfalik:
Closing to make it 1 commit.

PR #427, opened 2015-03-22
Update

G61:


G61:
Fuck this, this is an error.

CaptainElliott:
Personally I prefer having the color of the channel being red, also this needs to be redone to use localization

PR #450, opened 2015-03-23
Add Bukkit flavor, and version check

ryanw-ny:
I wrote this little tidbit of code for another project, and added it into CardinalPGM. Your more than welcome to use it if you wish. It checks to make sure the user is using SportBukkit, and if they aren't it disables the plugin. If they are, it then goes on to check the version and if they aren't using SportBukkit 1.8 it will disable the plugin, and forward them to the wiki page to get more information about what CardinalPGM's current supported SportBukkit version might be.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/Cardinal.java
+++ b/src/main/java/in/twizmwaz/cardinal/Cardinal.java
@@ -96,6 +96,18 @@ public class Cardinal extends JavaPlugin {
     @Override
     public void onEnable() {
         instance = this;
+        if (Bukkit.getName().equals("SportBukkit")) {
+            if (!getServer().getClass().getPackage().getName().contains("v1_8_R")) {
+                getLogger().log(Level.SEVERE, "Failed to load because you're running a invalid version of SportBukkit, disabling plugin...");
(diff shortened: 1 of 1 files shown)


twizmwazin:
We require SportBukkit 1_8_R1 now, as we use NMS.

ryanw-ny:
@twizmwazin It was suggested by @CaptainElliott in my last pull request to not hard code that in, since other revisions might be added. Does NMS break every revision?

CaptainElliott:
I guess we could keep the version updated

ryanw-ny:
@CaptainElliott I am aware of NMS breaking sometimes, during each Minecraft update, but not every revision, please correct me if I am wrong in this observation. Feel free to pull this if you wish @twizmwazin

twizmwazin:
@ryanw-ny Let me know when you have checked this. When NMS changes, the Minecraft version string changes. For example, 1.8.3 (with new NMS) is 1_8_R2. Any versions supporting the same NMS will keep the same version string.

ryanw-ny:
@twizmwazin I will test it as soon as I get home, and hard code that version in.

PR #453, opened 2015-03-23
Removed compiling section

TiTo-418:
Now can download the .jar


CaptainElliott:
You can't download a jar yet, and developers still need to compile

TiTo-418:
http://ci.twizmwaz.in/job/Cardinal/lastSuccessfulBuild/

CaptainElliott:
Developers need to compile their code to test it

PR #454, opened 2015-03-23
Update es.xml with the new translations

Eudaldca:
I also noticed there is a duplicate, noRecentPM and NoMessages, which both have the same translation (one was missing)


twizmwazin:
Thanks

PR #457, opened 2015-03-24
Added the new way of warns

vYuri:


iEli2tyree011:
Your build failed.

AngelCasillas01:
@vYuri Try instead of `ChatColorDARK_RED` use `ChatColor.DARK_RED` and  instead of `ChatColor.RED`

AngelCasillas01:
@iEli2tyree011 is that well? (^^)

Alfalik:
Use add instead of added. And, the actual warn message is identic to OCN, I dont see the reason to change it.

CaptainElliott:
The build failed, and what you was trying to do is pointless

PR #461, opened 2015-03-24
implement repairable monuments

### LINKED ISSUE #449: Add a feature which monuments can't be fixed.
In some XML maps (such as Revenge TE) On the XML file theres is not `<itemremove>` on monuments blocks (for example coal monument) and if you fixed that block counts like had never broke it

CaptainElliott:
forgot to put it in the commit message:

fixes #449


iEli2tyree011:
cool

PR #463, opened 2015-03-25
Fix(:D) command usage

TiTo-418:


iEli2tyree011:
The argument for /start isn't a team name, it's a time. If you can make another pull request where you replace the [time] with <time> for /start and <team> for /end, that would be appreciated.

TiTo-418:
Sorry! My error!

PR #479, opened 2015-03-27
Implement update notifications

CaptainElliott:
By editing the version in the `update` file it will send a notification that there is a new version available

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/Cardinal.java
+++ b/src/main/java/in/twizmwaz/cardinal/Cardinal.java
@@ -32,7 +32,6 @@ public class Cardinal extends JavaPlugin {
     private static LocaleHandler localeHandler;
     private CommandsManager<CommandSender> commands;
     private static Database database;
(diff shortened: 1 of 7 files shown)


twizmwazin:
I think this feature should be implemented using git commit hashes. I'm not sure if either Jenkins or GitHub has an api to hook into.

SungMatt:
@twizmwazin Sending a GET request to `https://api.github.com/repos/twizmwazin/CardinalPGM/git/refs/head/master` will return with the URL of the most recent commit (Under object). The returned URL should look like `https://api.github.com/repos/twizmwazin/CardinalPGM/git/commits/(SHA of commit)`

Alternatively, under object of the result of the first request, it returns SHA of the commit.

CaptainElliott:
I think it is good though that we can send a customized message with the update, stating what it is about or if it is an urgent bug fix

CaptainElliott:
re-writing with new implementation

PR #483, opened 2015-03-28
add mods use `/pardon` and `-f`

AngelCasillas01:


RubeGamerV:
Cardinal don´t have the command /pardon, this is implemented on SB

PR #500, opened 2015-03-30
Implement update notifications

### LINKED ISSUE #497: Update Notifications
Almost done, just need to do the JSON parsing

CaptainElliott:
This file currently needs to be created https://github.com/CaptainElliott/CardinalPGM/commit/ff0362058597321c0c614f7a1d6b097b4fb590d1#diff-f0e9d0bcff46f7f9cccd4180b4610642R18

following the same schema as https://raw.githubusercontent.com/CaptainElliott/test-ignore/master/update.json

It works by checking if the git version of the plugin is the same as the one in this repository, if not it gets the information from that file

closes #497

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/UpdateHandler.java
+++ b/src/main/java/in/twizmwaz/cardinal/UpdateHandler.java
@@ -1,0 +1,112 @@
+
+package in.twizmwaz.cardinal;
+
+import in.twizmwaz.cardinal.util.ChatUtils;
+import in.twizmwaz.cardinal.util.GitUtils;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.entity.Player;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class UpdateHandler {
+
+    private final String updateLocation = "https://raw.githubusercontent.com/twizmwazin/CardinalNotifications/master/update.txt";
+
+    private boolean update = false;
+
+    private String link;
+    private boolean urgent;
+    private String message;
+
+    public UpdateHandler() throws IOException {
+        BufferedReader in = new BufferedReader(new InputStreamReader(getUpdateURL().openStream()));
+        setLink(in.readLine().replace("link: ", ""));
+        setUrgent(Boolean.parseBoolean(in.readLine().replace("urgent: ", "")));
+        setMessage(in.readLine().replace("message: ", ""));
+
+    }
+
+    public void sendUpdateMessage(Player player) {
+        player.sendMessage(ChatColor.DARK_RED + "" + ChatColor.STRIKETHROUGH + "----------------" + ChatColor.BOLD + "" + ChatColor.GOLD + " Cardinal Update " + ChatColor.WHITE + "(" + GitUtils.getLatestGitRevision().substring(0, 6) + ")" + ChatColor.DARK_RED + " " + ChatColor.STRIKETHROUGH + "----------------");
+        player.sendMessage(ChatColor.DARK_RED + "" + ChatColor.BOLD + "Urgent: " + (urgent ? ChatColor.RED : ChatColor.GREEN) + urgent);
+        player.sendMessage(ChatColor.GOLD + link);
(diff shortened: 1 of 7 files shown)


src/main/java/in/twizmwaz/cardinal/util/GitUtils.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.gson.JsonParser;
+
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.URL;
+
+public class GitUtils {
+
+    public static String getLatestGitRevision(){
```
twizmwazin: Missing a space

twizmwazin:
- The update message should just be a Minecraft Json, which can be sent directly to the player. This would allow the "update" notification to be used outside of a specific format. We don't need the other fields, they can be included in the message field.
- The git revision should be stored as-is in the plugin, storing it just as a plain UTF-8 file in the jar would be sufficient. I don't like the idea of extracting it from the version string.

CaptainElliott:
Ah, I get what you mean by minecraft json now :p

twizmwazin:
@CaptainElliott Just make a file with ${describe} in the resources folder, it will get filtered by maven.

CaptainElliott:
@twizmwazin could you create the repo with something like this inside 
https://raw.githubusercontent.com/CaptainElliott/test-ignore/master/update.json

twizmwazin:
@CaptainElliott I'll do it as soon as I get home, at school. I'll be home approx 4:15 my time.

ryanw-ny:
@CaptainElliott I would wrap the SportBukkit server itself with a launch application, that can handle doing the updates, and other things. You could also have it automatically download SportBukkit, and have complete control over using the latest versions of things, etc.

CaptainElliott:
@ryanw-ny The Git API is being used and we have talked about having cardinal auto update itself in the future, not sure about SportBukit, but I don't think that needs to be done.

There's no harm in setting it to false :package: 

@twizmwazin ${describe} will return the whole version including artifact ID and only the short version of the commit, which is how I did it originally

CaptainElliott:
https://raw.githubusercontent.com/CaptainElliott/test-ignore/master/update.json Looks like

src/main/java/in/twizmwaz/cardinal/UpdateHandler.java
```diff
@@ -0,0 +1,73 @@
...
+        this.message = GitUtils.getUpdateMessage(updateLocation);
+        this.localGitRevision = GameHandler.getGameHandler().getPlugin().getDescription().getVersion().substring(GameHandler.getGameHandler().getPlugin().getDescription().getVersion().length() - 6, GameHandler.getGameHandler().getPlugin().getDescription().getVersion().length());
+    }
+
+    public void sendUpdateMessage(Player player) {
+        IChatBaseComponent chat = ChatSerializer.a(message);
+        PacketPlayOutChat packet = new PacketPlayOutChat(chat);
+
+        EntityPlayer nmsPlayer = ((CraftPlayer) player).getHandle();
+        nmsPlayer.playerConnection.sendPacket(packet);
```
twizmwazin: The NMS is unnecessary, use player#sendRawMessage(String json)

src/main/java/in/twizmwaz/cardinal/module/modules/updateNotification/UpdateNotification.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public void unload() {
+        HandlerList.unregisterAll(this);
+    }
+
+    @EventHandler
+    public void onPlayerJoin(PlayerJoinEvent event) {
+        if (event.getPlayer().isOp()) {
+            try {
+                updateHandler = new UpdateHandler();
```
twizmwazin: Why can't we just make one instance and use that?

src/main/java/in/twizmwaz/cardinal/UpdateHandler.java
CaptainElliott: sendRawMessage is not the same as /tellraw

dentmaged: Can't you deserialise it then use player#sendMessage(String message)?

twizmwazin:
Commits pushed manually

PR #518, opened 2015-04-03
Added Chinese Translation

SungMatt:
This chinese translation is written in a slightly informal manner, wherein some words are omitted, for example "你的队" is changed to "你队" as they mean the same, but this extra "的" is only used in a more formal setting. The level of chinese used for this is at conversational level between friends, not talking to elders or the president.

A few other words of note:

I have used some of these in certain context, like the `<of>` one was pretty ambigious, so I assumed that it is used in the case of  "Page 4  of 17".

This translation was based on its english counterpart created on the 6th of March. New tags might have been added, please leave me a message telling me which new tags to add.

A bit of information on myself:

Chinese is not my dominant language, but it is one of my two master languages (the other being English.) In my country we have to learn 2 languages, English and a Mother Tongue (So if you're Malay you learn Malay, being Chinese you learn Chinese, Indians learn Tamil/Hindi/other NTIL(Non-Tamil Indian Language)). I do believe that my level of Chinese is a level that is enough to read and write a short story, and I can use it fluently.

I have also checked this translation with the Chinese scholars from China, and they have read it and said it was smooth.

Other than that, I hope this will help.

### DIFF
--- a/src/main/resources/lang/zh.xml
+++ b/src/main/resources/lang/zh.xml
@@ -1,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<locale lang="zh">
+    <error>
+        <restartDuringMatch>在比赛中，不可以重启比赛。请用-f去强制刷新。</restartDuringMatch>
+        <cycleDuringMatch>在比赛中，不可以换游戏地图。请用-f去强制。</cycleDuringMatch>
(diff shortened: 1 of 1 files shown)


twizmwazin:
Thanks

SungMatt:
You're welcome, @twizmwazin .

PR #530, opened 2015-04-03
Add /play and /begin commands.

Vixo1:
A bit useless, but it is on Overcast's PGM.


iEli2tyree011:
Can you make this one commit please?

PR #535, opened 2015-04-04
Make HTML file toggleable - enabled by default

dentmaged:
Nothing much. All in the title.


CaptainElliott:
IMO it should still generate the statistic HTML file and the log file just not upload them if the config is set to false

dentmaged:
Okay! I'll make a new pull request!

PR #538, opened 2015-04-04
Update LICENSE

ItzWid13:


SungMatt:
The original version is the correct way of writing the MIT license.

PR #540, opened 2015-04-04
Update es.xml

ItzWid13:


Alfalik:
"La", at the start of the message, means singular, not plural.

PR #546, opened 2015-04-05
Toggle uploading and milliseconds in transcript

dentmaged:
All in the title!

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/stats/Stats.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/stats/Stats.java
@@ -1,7 +1,6 @@
 package in.twizmwaz.cardinal.module.modules.stats;
 
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
+import in.twizmwaz.cardinal.Cardinal;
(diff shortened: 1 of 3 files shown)


src/main/java/in/twizmwaz/cardinal/module/modules/stats/Stats.java
```diff
@@ -34,15 +44,11 @@
...
-import org.jsoup.nodes.Attributes;
 import org.jsoup.nodes.Document;
 import org.jsoup.nodes.Element;
-import org.jsoup.parser.Tag;
 
-import java.io.*;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
+import com.google.common.collect.Lists;
```
iEli2tyree011: Why did you import google commons?

iEli2tyree011:
Thanks

src/main/java/in/twizmwaz/cardinal/module/modules/stats/Stats.java
dentmaged: Ermmmm. That is weird. Sorry!

PR #552, opened 2015-04-05
Update config.yml

Zenzoh:
Some things I think should be modified


twizmwazin:
You actually have change the plugin it self to change the config file.

PR #557, opened 2015-04-06
Add shutdown after x matches

dentmaged:
I don't know the server restarting colours so I guessed them based on cycling.

Please make a PR if someone knows!


CaptainElliott:
I don't think this should be implemented as we don't have a restart script and neither will other servers

twizmwazin:
As Elliott said, this isn't necessary and would likely be found to be a nuisance.

PR #560, opened 2015-04-06
Add auto potion bottle remover

ShinyDialga45:


ShinyDialga45:
Thanks, focused too much on making the module itself good :P

iEli2tyree011:
Can you make this into one commit please?

iEli2tyree011:
Also, I'd recommend storing the slot value as a `final int` and then referring to it in the runnable, as the player may have switched items that they are holding.

iEli2tyree011:
Looks good. Thanks!

PR #578, opened 2015-04-11
Add /say & /broadcast commands

Alfalik:
There we go. Added localized messages and register the command.


iEli2tyree011:
Merging, but there's still quite a few problems. I'll fix them up.

Alfalik:
Uh, thanks

PR #582, opened 2015-04-12
added default channels on es.xml

AngelCasillas01:
Checked on apple's keyboard, everything should be fine 🙌🏽


Superjd10:
Ha*

Ha: Haber (have).
A: Preposición (preposition).

Pro tip: Never trust Apple's keyboard. Never trust an auto-corrector, translate from what you really know.

AngelCasillas01:
@Superjd10 So sorry, making a new pull

PR #589, opened 2015-04-13
Add observer book - updated!

dentmaged:
I've updated it to use a better technique and it doesn't require a module!


dentmaged:
@twizmwazin What about this?

dentmaged:
Closing - see #602

PR #594, opened 2015-04-14
Add configuration option to kill mobs on map load

dentmaged:
When editing maps sometimes mob spawn. I think this would be a useful feature for those who don't use MCEdit, however I don't think that this is a requirement. I will understand if you don't want to merge it.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
@@ -1,6 +1,7 @@
 package in.twizmwaz.cardinal.module;
 
 import in.twizmwaz.cardinal.match.Match;
+import in.twizmwaz.cardinal.module.modules.antimob.AntiMobBuilder;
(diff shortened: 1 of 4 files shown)


AngelCasillas01:
Try to add this but for kill mobs within 2 blocks because some maps like Cake Wars there is spiders

dentmaged:
@AngelCasillas01 What do you mean by "kill mobs within 2 blocks"? 2 blocks within what? a region?

CoWinkKeyDinkInc:
@dentmaged Paintings, itemframes, and the player are also entities that you'd may want to keep in a map.

dentmaged:
@CoWinkKeyDinkInc Of course! Thanks @twizmwazin and @CoWinkKeyDinkInc!

CaptainElliott:
This is a map issue, not a plugin issue. Maps may want to have things spawned in them inside the world files, it is something we should not allow users to disable.

dentmaged:
@CaptainElliott Okay! It's disabled by default. Maybe something in the XML, although it wouldn't work on Overcast, it would be good to test maps with.

CaptainElliott:
Overcast has a mobs filter

PR #613, opened 2015-04-18
Update es.xml

Eudaldca:


TiTi-418:
disagree with line 123

Superjd10:
I agree with @TiTi-418

Eudaldca:
I have just seen it. I translated from the english string without looking at the text between "<>". In overcast, the max-overfill refers to the amount of premium users that can join when teams are full, so in the english translation of cardinal it should be written tesm size indtead of max-overfill. Feel free to make a new PR with "máxima capacidad", I agree with you. Make it also for en.xml.

TiTo-418:
@Superjd10 Can you make it in en.xml?

Superjd10:
@TiTo-418 what do you mean?

PR #629, opened 2015-04-29
Improve proximity, score and time limit commands

Alfalik:
Here's a video showing the time limit changes: https://youtu.be/52too94df7w
Besides, the console now can use ```/score```, and the proximity command show the wool message as its respective color (http://gyazo.com/66e8ae6ad19a8ba98a1d7759f0f47aed).

**Edit**: @iEli2tyree011 Already fix some errors, check again. Thanks

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/command/TimeLimitCommand.java
+++ b/src/main/java/in/twizmwaz/cardinal/command/TimeLimitCommand.java
@@ -4,6 +4,9 @@ import com.sk89q.minecraft.util.commands.*;
 import in.twizmwaz.cardinal.GameHandler;
 import in.twizmwaz.cardinal.chat.UnlocalizedChatMessage;
 import in.twizmwaz.cardinal.event.TimeLimitChangeEvent;
+import in.twizmwaz.cardinal.match.Match;
+import in.twizmwaz.cardinal.match.MatchState;
+import in.twizmwaz.cardinal.module.modules.matchTimer.MatchTimer;
 import in.twizmwaz.cardinal.module.modules.timeLimit.TimeLimit;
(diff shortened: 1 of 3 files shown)


src/main/java/in/twizmwaz/cardinal/command/TimeLimitCommand.java
```diff
@@ -18,8 +21,11 @@
     @Command(aliases = {"timelimit", "tl"}, desc = "Modify the time limit of the current match.", usage = "<add, set> <time> [result]")
     public static void timeLimit(final CommandContext cmd, CommandSender sender) throws CommandException {
         if (cmd.argsLength() == 0) {
-            if (TimeLimit.getMatchTimeLimit() != 0) {
+            Match match = GameHandler.getGameHandler().getMatch();
+            if ((TimeLimit.getMatchTimeLimit() != 0) && (!(match.getState().equals(MatchState.ENDED)) || (match.getState().equals(MatchState.CYCLING)))) {
```
iEli2tyree011: What is the reason for this, and did you make sure it worked?

Alfalik: @iEli2tyree011 If a player do ```/timelimit``` when the match is starting or running, it will say the actual time limit, on the other hand, if the command has been executed when the match is cycling or end, the message will report the time that last the match. Already tested, but I'm very busy to test again right now or send you an image, sorry. Although, you can test it

dentmaged: @Alfalik You won't be able to run the command once the game has ended.

PR #644, opened 2015-05-10
Change Team command

dentmaged:
I changed the team command to use nested commands instead.


src/main/java/in/twizmwaz/cardinal/command/TeamCommands.java
```diff
@@ -0,0 +1,142 @@
...
+import com.sk89q.minecraft.util.commands.CommandUsageException;
+import com.sk89q.minecraft.util.commands.NestedCommand;
+
+public class TeamCommands {
+
+    public static class TeamParentCommand {
+        @Command(
+                aliases = { "team" },
+                desc = "Manage the teams in the match.",
+                min = 1,
```
iEli2tyree011: Why'd you have a min and max here? You should be able to do a min and max specific to each nested command, right?

dentmaged: Oh, I'll fix that!

iEli2tyree011:
Thanks!

PR #651, opened 2015-05-18
Update ProximityCommand.java

TiTo-418:


Alfalik:
The closest kill is only for wools, not for cores and monuments.

PR #667, opened 2015-05-23
Use player's display name/player list name for rank prefix, but strip it of all stars and non-ASCII symbols

dentmaged:
This commit allows for players to use /nick (in Essentials and many other plugins) and allow the user to still have a prefix. This tactic replaces all non-ASCII symbols with nothing (thus replacing the admin star - ❖. This system then replaces all asterisks as well).


dentmaged:
Any opinions?

twizmwazin:
@iEli2tyree011

iEli2tyree011:
I'm still not a fan of this solution. I told you about the Sportbukkit "fake name" API, did you try that yet?

dentmaged:
@iEli2tyree011 I tried it, but it didn't work. I used ```                player.setFakeName(bukkit, nick);
                player.setFakeDisplayName(bukkit, nick);```

twizmwazin:
@dentmaged I'm going to take a shot in the dark here and say the issue was with your code, not sportbukkit.

dentmaged:
@twizmwazin Yeah, but I don't see what's wrong with my code.

TiTo-418:
@dentmaged This PR should be rebased

PR #672, opened 2015-05-26
Update the player's inventory when it is reset

dentmaged:
The player's inventory should automatically be updated, however, it is a bug in Bukkit. ```player.updateInventory();``` is a temporary fix, hence it is deprecated. However, it seems that it hasn't been fixed.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/observers/ObserverModule.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/observers/ObserverModule.java
@@ -16,6 +16,7 @@ import in.twizmwaz.cardinal.module.modules.team.TeamModule;
 import in.twizmwaz.cardinal.module.modules.tutorial.Tutorial;
 import in.twizmwaz.cardinal.util.ItemUtils;
 import in.twizmwaz.cardinal.util.TeamUtils;
+
 import org.bukkit.*;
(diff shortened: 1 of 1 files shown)


iEli2tyree011:
Why should we add it if it doesn't fix anything?

dentmaged:
@iEli2tyree011 When a game ends, you get ghost items in your inventory, in order to update that slot, you have to right click. Sorry for being unclear, but by me saying ```However, it seems that it hasn't been fixed.```, I was refering to the bug in Bukkit (about the inventory not being updated automatically).

iEli2tyree011:
@dentmaged Okay. Do you mind also including it after the other places where the player's inventory is changed, then?

dentmaged:
@iEli2tyree011 This is the only place where the inventory needs to be updated.

iEli2tyree011:
@dentmaged It appears that the player's inventory is also changed in the CardinalSpawnEvent.

dentmaged:
@iEli2tyree011 The issue only occurs once a game has ended, the player still has his items (but in the form of a ghost item).

TiTo-418:
http://youtu.be/867sIdBh_uU

dentmaged:
Thank you @TiTo-418. This PR fixes this issue.

twizmwazin:
Merged

PR #674, opened 2015-05-28
[WIP] Add sounds and change how proximity calculates a touch distance

Alfalik:
+ In ad507af I added toggleables sounds for warns & monument modes.
+ Now, in the 2bdea51 commit, the ```location``` is only needed to calculate the distance when a player  kill an enemy. In the case of a touch, proximity should get the distance between the player-monument, not the location-monument.
+ I was planing to add a request command, here's a gist to see how it is: https://gist.github.com/Alfalik/be1dfa2ba355f4464bcf

In addition, I'm just seeing if the code is correctly write and if it is, convert all the code in one commit.


TiTo-418:
Deberias hacer esto en una commit para que lo mergeen / You need to do this in one commit to merge

Alfalik:
@TiTo-418 Read first, please.

Prodigy4532:
Nice.

Alfalik:
Oops! Wrong commands with git bash

PR #686, opened 2015-06-02
Deny executing /me during a global mute, fixes #680

### LINKED ISSUE #680: Global mute lets you execute /me
When in global mute, you can still do /me. It was fixed for the single mute, but not for the gmute.

Also, since its name is *global* mute, shouldn't it only affect global chat, rather than both team and global?

ryanw-se:
Fixes bug #680


twizmwazin:
Thanks

PR #690, opened 2015-06-09
Check if the player's team is null in Playable

### LINKED ISSUE #689: Rebase to 1.8.7
Cardinal doesn't work on SportBukkit 1.8.7

dentmaged:
This half fixes #689. Tthe issue still occurs, it just doesn't print a NullPointerException.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/playable/Playable.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/playable/Playable.java
@@ -29,7 +29,7 @@ public class Playable implements Module {
 
     @EventHandler
     public void onPlayerMove(PlayerMoveEvent event) {
-        if (GameHandler.getGameHandler().getMatch().isRunning() && !TeamUtils.getTeamByPlayer(event.getPlayer()).isObserver()) {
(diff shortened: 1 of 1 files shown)


twizmwazin:
When would a player's team be null? This could only occur if there is a plugin issue elsewhere, in which case it would be a higher priority to fix that.

dentmaged:
@twizmwazin There are always checks like that and look at the error in #689.

twizmwazin:
@dentmaged That is not an issue with Playable, however, it is the server using a mismatching version of SportBukkit.

dentmaged:
@twizmwazin there was an NPE pointing to line 32 in Playable. My only guess is that the player's team is null or the match is null.

twizmwazin:
@dentmaged The player's team was null, however the issue isn't within this class, but the fact that the player wasn't assigned a team.

iEli2tyree011:
I'm fine with merging this, I personally always check for teams being null just in case. If this shouldn't be merged, then perhaps we should remove the other checks for consistency, as @dentmaged was saying earlier. @twizmwazin

twizmwazin:
I have a ton of local changes that I am hoping can be pushed tonight to a new branch replacing the team methods with Optional<TeamModule> methods. This will force a check.

dentmaged:
@twizmwazin Awesome! Will the other branch be a "develop" styled branch?

iEli2tyree011:
@twizmwazin Are you against merging this now, however? We may as well merge this to master to keep the code consistent as the new teams system is being developed.

twizmwazin:
sure

PR #708, opened 2015-06-19
Title Respawn

TheMolkaPL:
This pull request will add a new death and respawn system from the Overcast Network. The documentation of this module is available [on the docs page](https://docs.oc.tc/modules/spawns#respawn). This pull will close issue #699 too.

I have a few questions;
* how should we send titles?
* how should we prevert dead people to interact with the gameplay (exactly as the observers)?
* how `blackout` should work?

**EDIT** I known about getOrDefault, I've fixed it.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java
@@ -1,0 +1,32 @@
+package in.twizmwaz.cardinal.module.modules.titleRespawn;
+
+import in.twizmwaz.cardinal.match.Match;
+import in.twizmwaz.cardinal.module.BuilderData;
+import in.twizmwaz.cardinal.module.ModuleBuilder;
+import in.twizmwaz.cardinal.module.ModuleCollection;
+import in.twizmwaz.cardinal.module.ModuleLoadTime;
+import in.twizmwaz.cardinal.util.StringUtils;
+import org.jdom2.Element;
+
+@BuilderData(load = ModuleLoadTime.NORMAL)
+public class TitleRespawnBuilder implements ModuleBuilder {
+
+    @Override
+    public ModuleCollection load(Match match) {
+        Element head = match.getDocument().getRootElement().getChild("respawn");
+
+        int delay = StringUtils.timeStringToSeconds(head.getAttributeValue("delay", "2s"));
+        if (delay < 1) {
+            delay = 1;
+        }
+
+        boolean auto = Boolean.parseBoolean(head.getAttributeValue("auto", "false"));
+        boolean blackout = Boolean.parseBoolean(head.getAttributeValue("blackout", "false"));
+        boolean spectate = Boolean.parseBoolean(head.getAttributeValue("spectate", "false"));
+        boolean bed = Boolean.parseBoolean(head.getAttributeValue("bed", "false"));
+
+        TitleRespawn module = new TitleRespawn(delay, auto, blackout, spectate, bed);
+        return new ModuleCollection<>(module);
+    }
+
+}
--- a/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
@@ -64,6 +64,7 @@ import in.twizmwaz.cardinal.module.modules.teamPicker.TeamPickerBuilder;
 import in.twizmwaz.cardinal.module.modules.timeLimit.TimeLimitBuilder;
 import in.twizmwaz.cardinal.module.modules.timeLock.TimeLockBuilder;
 import in.twizmwaz.cardinal.module.modules.timeNotifications.TimeNotificationsBuilder;
+import in.twizmwaz.cardinal.module.modules.titleRespawn.TitleRespawnBuilder;
 import in.twizmwaz.cardinal.module.modules.tnt.TntBuilder;
 import in.twizmwaz.cardinal.module.modules.tntDefuse.TntDefuseBuilder;
 import in.twizmwaz.cardinal.module.modules.tntTracker.TntTrackerBuilder;
@@ -160,7 +161,8 @@ public class ModuleFactory {
             UpdateNotificationBuilder.class,
             MatchTranscriptBuilder.class,
             PotionRemoverBuilder.class,
-            InvisibleBlockBuilder.class
+            InvisibleBlockBuilder.class,
+            TitleRespawnBuilder.class
     };
     private final List<ModuleBuilder> builders;
 
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawn.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawn.java
@@ -1,0 +1,212 @@
+package in.twizmwaz.cardinal.module.modules.titleRespawn;
+
+import in.twizmwaz.cardinal.GameHandler;
+import in.twizmwaz.cardinal.event.MatchStartEvent;
+import in.twizmwaz.cardinal.module.TaskedModule;
+import in.twizmwaz.cardinal.util.PlayerUtils;
+import java.util.Arrays;
+import java.util.HashMap;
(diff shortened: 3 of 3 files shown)


dentmaged:
Nice. You can send titles, I'll PR your repo later. Make sure you prevent the user from dismounting from the horse. In Title Respawn, will you remove the comments if not, there are spelling mistakes on line 147 (broke should be break), line 119 (reseted should be reset).

twizmwazin:
For titles, they can be sent using SportBukkit API. The relevant patch is [here] (https://github.com/OvercastNetwork/SportBukkit/blob/master/Bukkit/0034-Title-API.patch).

For dead players, in the observers module there could be a check for if the player is dead, that seems fine to me.

Blackout is just a blindness potion effect, I'm not too sure what you are asking.

src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java
```diff
@@ -0,0 +1,32 @@
...
+    @Override
+    public ModuleCollection load(Match match) {
+        Element head = match.getDocument().getRootElement().getChild("respawn");
+
+        int delay = StringUtils.timeStringToSeconds(head.getAttributeValue("delay", "2s"));
+        if (delay < 1) {
+            delay = 1;
+        }
+
+        boolean auto = Boolean.parseBoolean(head.getAttributeValue("auto", "false"));
```
twizmwazin: You may want to use NumUtils#parseBoolean, as it allows the XML to use `on` and `off` in addition to `true` and `false`.

TheMolkaPL:
Thanks for the answers!

> Make sure you prevent the user from dismounting from the horse.

Which event (?) I should use?

dentmaged:
@TheMolkaPL VehicleExitEvent

src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java
```diff
@@ -0,0 +1,32 @@
...
+import in.twizmwaz.cardinal.module.ModuleLoadTime;
+import in.twizmwaz.cardinal.util.StringUtils;
+import org.jdom2.Element;
+
+@BuilderData(load = ModuleLoadTime.NORMAL)
+public class TitleRespawnBuilder implements ModuleBuilder {
+
+    @Override
+    public ModuleCollection load(Match match) {
+        Element head = match.getDocument().getRootElement().getChild("respawn");
```
dentmaged: Shouldn't there be a check for pre-1.8 maps?

TheMolkaPL:
```java
if (head == null) {
    return new ModuleCollection(new TitleRespawn(2, false, false, false, false));
}
```
?

dentmaged: @TheMolkaPL
```ModuleCollection results = new ModuleCollection(); Element head = match.getDocument().getRootElement().getChild("respawn"); if (head == null) { return results; } // Rest of code here```

Comment (TheMolkaPL) [inline src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java]: But this module is enabled on every Overcast map; it's a default OCN respawn system now.

Comment (dentmaged) [inline src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java]: @TheMolkaPL Many people (such as myself) have still got old maps on Cardinal. I have 50+ maps on my server and I don't have enough time to update it all.

Comment (TheMolkaPL) [inline src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java]: @dentmaged but I need to use

```java
if (head == null) {
    return new ModuleCollection(new TitleRespawn(2, false, false, false, false));
}
```

because if the XML doesn't has a `<respawn />` tag the `head` will be `null`, so this https://github.com/TheMolkaPL/CardinalPGM/commit/8a4f0aa67cf24d4b276f8772480356a4c89604a5#diff-8989480d1f8f58581249e0dcac07d95bR23 will be a `NullPointerException`.

dentmaged:
@TheMolkaPL Yeah, I would have default values and check for NPEs

src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawnBuilder.java
dentmaged: @TheMolkaPL Yeah, I would have default values and check for NPEs

Pablete1234:
docs page:
https://docs.oc.tc/modules/spawns
the delay defaults to 2 seconds, and the minimum is 1 second
they should not need an update, they would use default values

TheMolkaPL:
@Pablete1234 2 seconds is the default value.

```java
StringUtils.timeStringToSeconds(head.getAttributeValue("delay", "2s"));
```

TheMolkaPL:
@twizmwazin @dentmaged what do you think about two new events:

* `CardinalRespawnEvent` - called when player is able to respawn (after the scheduling)
* `CardinalRespawnClickEvent` - (cancellable) called when player left-clicks and tries to respawn.

May be useful for the CTF mode or other modules.

twizmwazin:
@TheMolkaPL Cardinal already has it's own `[CardinalSpawnEvent](https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/event/CardinalSpawnEvent.java)` and `[CardinalDeathEvent]{https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/event/CardinalDeathEvent.java).

twizmwazin:
Cardinal already has a CardinalSpawnEvent and CardinalDeathEvent

TheMolkaPL:
But `CardinalSpawnEvent` is called when player respawns. I will call `CardinalRespawnEvent` when player is able to respawn. In other words; when scheduling to respawn "Left click to respawn in Xs" are "Left click to respawn". The subtitle of the respawn screen may be change to for example "You will respawn when the flag is dropped" istead of "Left click to respawn".

twizmwazin:
Wouldn't the module's titles be controlled from within the module? Is there any use in Cardinal for this event? I'm not against adding it, although it would be more appropriate to name it CardinalPlayerAbleToRespawnEvent

dentmaged:
@twizmwazin I use Cardinal in other plugins so I can keep updating Cardinal and still have my features. I think providing an API for developers is a good idea. Your choice though.

twizmwazin:
@dentmaged What 'features' have you implemented?

dentmaged:
@twizmwazin Autostart because I'm too lazy to type ``/join blue``` and ```/start 3```

Pablete1234:
@twizmwazin Or @TheMolkaPL Remove the [WIP] thing in the title, it has a label

TiTo-418:
@Pablete1234 It is the same .-.

dentmaged:
@TiTo-418 What? It has a label, it doesn't need ```[WIP]``` in the title anymore.

Pablete1234:
@TheMolkaPL do not make the player ride a horse, make it ride an armor stand, you can give it invisible:1b, and marker:1b, marker tag makes no collision box (you can place blocks inside it), and with invisible no one can see it

dentmaged:
@TheMolkaPL Are you going to finish this?

TheMolkaPL:
@dentmaged I don't have access to my computer right now.

dentmaged:
@TheMolkaPL I see. Sorry.

dentmaged:
@TheMolkaPL Why are you closing this? Will this be rebased or are you unable to continue it?

TheMolkaPL:
I don't have time to continue this feature. I can push my local changes if you want.

Pablete1234:
@TheMolkaPL push your local changes, they are always going to be userfull for whoever keeps implementing this

TheMolkaPL:
https://github.com/TheMolkaPL/CardinalPGM/commit/d68abeb11182a22b948a39093f162df4fb5b1243

PR #713, opened 2015-06-23
Only move the rotation if the next map is the next map in the rotation

dentmaged:
Used on Overcast. It may not be the most useful thing for scrimmages, but I like having it on my server.

This PR will only move the position in the rotation if the next map is the next map in the rotation.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/GameHandler.java
+++ b/src/main/java/in/twizmwaz/cardinal/GameHandler.java
@@ -44,7 +44,9 @@ public class GameHandler {
     }
 
     public void cycleAndMakeMatch() {
-        rotation.move();
+        if (rotation.getNext().equals(cycle.getMap())) {
+            rotation.move();            
+        }
         World oldMatchWorld = matchWorld == null ? null : matchWorld.get();
         cycle.run();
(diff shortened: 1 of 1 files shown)


twizmwazin:
If you can fix that one small thing, I think this is fine to merge. @iEli2tyree011 if you want to check

iEli2tyree011:
Looks fine. But I agree, you shouldn't check the map by name, but instead by the actual object.

dentmaged:
@twizmwazin @iEli2tyree011 Done! :)

Pablete1234:
this makes the rot useless ??

dentmaged:
@Pablete1234 No, it only moves the rotation's position if the next map to be cycled is the next map in the rotation (much like OCN has). This works and I have tested it on my server.

Pablete1234:
ah ok, was more like a bug that it passed the index in the rot when it shouldnt

when i first read it i thougt it kinda disabled rot and only cycled if you did /sn first

dentmaged:
@twizmwazin Done! :)

dentmaged:
@iEli2tyree011 Thanks!

twizmwazin:
Merged

PR #726, opened 2015-06-24
[WIP] Add sounds, request command, auto-whitelist toggleable system and some more functions

Alfalik:
+ Request command for simples request like whitelist players or kick someone.
+ Change how proximity calculates victory monument-safeties vector (in discussion)
+ Sounds:
    - Monument modes
    - Warns
+ More usages for global mute command
+ Message for team change name is now send to global chat.
+ Send titles for match start and end (need anyone to implement titles)
+ Change colors and clarify some messages.

Waiting for opinions

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
@@ -1,5 +1,6 @@
 package in.twizmwaz.cardinal.module.modules.startTimer;
 
+import in.twizmwaz.cardinal.Cardinal;
 import in.twizmwaz.cardinal.GameHandler;
 import in.twizmwaz.cardinal.chat.ChatConstant;
 import in.twizmwaz.cardinal.chat.LocalizedChatMessage;
@@ -15,6 +16,7 @@ import in.twizmwaz.cardinal.util.ChatUtils;
 import in.twizmwaz.cardinal.util.TeamUtils;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
+import org.bukkit.OfflinePlayer;
 import org.bukkit.Sound;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Cancellable;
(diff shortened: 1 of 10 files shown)


src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
```diff
@@ -74,6 +89,22 @@ public void run() {
                     }
                 }
             }
+            for (Player player : Bukkit.getOnlinePlayers()) {
+                if (!TeamUtils.getTeamByPlayer(player).isObserver()) {
+                    if (time == 3) {
```
iEli2tyree011: The `time` variable keeps track of the countdown in ticks, not seconds.

src/main/java/in/twizmwaz/cardinal/module/modules/wools/WoolObjective.java
```diff
@@ -145,11 +145,7 @@ public void onWoolPickup(InventoryClickEvent event) {
                         this.touched = true;
                         if (touchMessage) {
                             double newProx;
-                            if (location != null) {
-                                newProx = location.distance(place.getVector());
-                            } else {
-                                newProx = player.getLocation().toVector().distance(place.getVector());
-                            }
+                            newProx = player.getLocation().toVector().distance(place.getVector());
```
iEli2tyree011: I understand what you're trying to fix with this now, but I think we should use the XML-specified location initially when picking up the wool, so perhaps we should separate out the upcoming `if` statement based on `!oldState`, when the first touch is achieved, and `newProx < proximity`, then the player's distance is less than the current proximity.

src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
dentmaged: You can also use

```java
if (time <= 60 && time != 0) {
    sendTitle(ChatColor.YELLOW + (time / 20));
} else if (time == 0) {
   sendTitle(ChatColor.GREEN + "Go!");
}
```

iEli2tyree011: @dentmaged That would send a message every tick. While it wouldn't show since you're dividing by an integer, meaning you'll get an integer (3, 2, 1), it's needless and could be made more efficient. Perhaps just put it in the `if` for the countdown, which only executes every time the time is divisible by 20 (one second).

Alfalik: @iEli2tyree011 Already tried this, it is correct?

```java
if (time == 60 || time == 40 || time == 20) {
    sendTitle.(ChatColor.YELLOW + (time / 20))
}
if (time == 0) {
    sendTitle.(ChatColor.GREEN + "Go!")
}
```

src/main/java/in/twizmwaz/cardinal/module/modules/wools/WoolObjective.java
Alfalik: @iEli2tyree011 Made some changes, now if the players pick up the wool clicking it in a chest (that should be the first touch) proximity will use location XML to calculate the distance. In the case that the wool is in the floor or dropped, proximity will use the distance between player-monument to calculate the vector. That's what do you want?

src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
CaptainElliott: @Alfalik use the modulus function to check that it is divisible by 20 instead of checking them manually

twizmwazin:
```java
if (time == 0) {
    sendTitle(ChatColor.GREEN + "Go!")
} else if (time %% 20 == 0) {
    sendTitle(ChatColor.YELLOW + (time / 20))
}
```

iEli2tyree011: @CaptainElliott @twizmwazin Modulus in Java is `%`. Not sure if `%%` has a different function.

src/main/java/in/twizmwaz/cardinal/module/modules/wools/WoolObjective.java
iEli2tyree011: @Alfalik No, the functions of picking up a wool off the ground and taking it out of a chest should be the same. In the case of maps like Dynamo, wool comes from a spawner and you have to pick it up.

Alfalik: @iEli2tyree011 Completely agree with you, but can't find a solution. Leaving it for you since you know more than me and understand the issue.

src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
```diff
@@ -74,6 +89,18 @@ public void run() {
                     }
                 }
             }
+            for (Player player : Bukkit.getOnlinePlayers()) {
+                if (!TeamUtils.getTeamByPlayer(player).isObserver()) {
+                    if (time == 0) {
+                        //sendTitle(ChatColor.GREEN + "Go!");
+                    } else if (((time % 20) == 0) && (time <= 60)) {
+                        //sendTitle(ChatColor.YELLOW + (time / 20));
+                    }
```
Alfalik: Already tested this sending messages to the chat, this is what I got. http://gyazo.com/c2f9579f5a523fb005533810169a398a

Alfalik:
I'll make this more tidy and well write.

PR #729, opened 2015-06-26
Update tab header colors for authors

Pablete1234:
this is how it is now on oc.tc
http://gyazo.com/28edae3d60fb814e74b9b04847ceadfb

PR #730, opened 2015-06-26
Change the PlayerList title

dentmaged:
This PR expands on #729. All it does is change to use the player's display name in the player list header.


iEli2tyree011:
If I'm not mistaken, contributors may not have a UUID. You would want to use their name in that case. Also, we may want to consider making the getName method part of a util class as it likely can have many other users.

CaptainElliott:
Yes, a UUID does not have to be defined, however if one is not defined then that name will not get any recognition so they can just be given a cyan name

dentmaged:
@iEli2tyree011 @CaptainElliott Done!

dentmaged:
@iEli2tyree011 @CaptainElliott Done!

CaptainElliott:
@dentmaged neat!

Pablete1234:
going to merge?

dentmaged:
@iEli2tyree011 Done!

iEli2tyree011:
@dentmaged Looks good, thanks!

PR #743, opened 2015-06-30
Implement an auto-whitelist system

Alfalik:
I have been learning some git commands and how to use branches. I'd prefer to have different pull requests and commit names since now, that's why I'm separating in "pieces" #726.

About this PR, it just delete the current players from the whitelist (at the start of every match) and whitelist all online players, also it turn on. And, when the match end, the whitelist is set to off. This feature is totally toggleable, as shown in the ```config.yml```.
Tell me if you have any thoughts about this implementation.

Here's an example of how it looks: http://gyazo.com/389a6efc92d76295f4c226be1a5aa3dc

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/chat/ChatConstant.java
+++ b/src/main/java/in/twizmwaz/cardinal/chat/ChatConstant.java
@@ -125,6 +125,8 @@ public enum ChatConstant {
     GENERIC_AUTO_BALANCE("generic.autoBalance"),
     GENERIC_YOU_CHANGED_RATING("generic.youChangedRating"),
     GENERIC_TEAM_SIZE_CHANGED("generic.teamSizeChanged"),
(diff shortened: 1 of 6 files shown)


dentmaged:
@Alfalik What about a team with players on it where not every player played. Wouldn't they be removed from the whitelist?

Alfalik:
@dentmaged Sorry, but can't understand what you mean. Could you write one example?

dentmaged:
@Alfalik I'm in the team Implosion. If my team is whitelisted and the game starts, they will be re-added. When the game ends, they'll be unwhitelisted.

TiTo-418:
@Alfalik This PR should be rebased

Alfalik:
@TiTo-418 If anyone can make an opinion, I'll rebase it, if not, no.

Alfalik:
Closing as this seems to be useless.

TiTo-418:
@Alfalik @twizmwazin Could we keep this opened?

PR #746, opened 2015-07-03
Update max-build-height msg

Pablete1234:
http://gyazo.com/d837728fdaf45b020f7803710dd679ba


twizmwazin:
Other than that, looks fine

Pablete1234:
first one isn't intended, second one is

PR #752, opened 2015-07-08
Support teams being optional

twizmwazin:
This pull request is here to ask others to test this branch, and find any bugs or inconsistencies before this I merge this into master. If you know what you are doing please feel free to do everything you can to try and break this, and report back the issues.


Pablete1234:
What's the purpose of "optional teams" what does that mean?

dentmaged:
@Pablete1234 What if the match is FFA? That's what Elly said to me when I asked about it.

Pablete1234:
i se FFA as 1 team with friendly fire on

twizmwazin:
@Pablete1234 That wouldn't support per-player scoring

Pablete1234:
@twizmwazin That makes sense

Eudaldca:
I found an issue, if you try to /end a regular match (Tested on boom TE) it gives out an error. [Log](http://pastebin.com/pRgxMh66/)
Also, if you try to cycle to an FFA match, it fails to cycle, [the xml](https://oc.tc/forums/topics/556b42175f35b90fe200689b/) is valid. I don't know if it is supposed to be able to cycle already, but I posted it just in case.

twizmwazin:
@Eudaldca Thanks, fixed

PR #750, opened 2015-07-06
Update join,start and end chat messages

Pablete1234:
More chat message updates (tested code)

PR #755, opened 2015-07-11
Update join, start, end, monument and core messages

Pablete1234:
Messed up some stuff with git in #750 , I reopen it here


twizmwazin:
This is going to need to be rebased on master before it can be merged

PR #762, opened 2015-07-13
Re-write cancel command

Alfalik:
This pull request just clarify what countdowns are cancelled, and also, make messages more clearer.


twizmwazin:
Looks good at first glance, @iEli2tyree011

Alfalik:
Oops! My mouse is the guilty

twizmwazin:
Few things from testing:
 - An exception is thrown if there is no timer to cancel.
 - You make the cancel messages look like warning messages, they aren't.
 - A wording thing, the map cycles, not the match.

Alfalik:
@twizmwazin First, thanks for testing. Now:

+ The exception has been solved.
+ Removed the warning
+ Changed the message, thanks for the correction

twizmwazin:
Going to close this as I don't think it is still relevant. Let me know if I am mistaken.

PR #764, opened 2015-07-14
Implement boss bars

Alfalik:
I've coded a basic boss bar for time remaining, here's a video of how it works and looks (https://youtu.be/X7fEoY2Mol0). I'll be thankful if someone help contributing to any of these inconsistencies:

+ Wither being visible sometimes
+ Bar flashing occasionally

If you find any problem(s), don't forget to reply here, thanks.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
@@ -2,7 +2,11 @@ package in.twizmwaz.cardinal.module.modules.bossBar;
 
 import com.google.common.collect.Maps;
 import in.twizmwaz.cardinal.Cardinal;
+import in.twizmwaz.cardinal.GameHandler;
 import in.twizmwaz.cardinal.chat.ChatMessage;
+import in.twizmwaz.cardinal.event.CycleCompleteEvent;
+import in.twizmwaz.cardinal.event.MatchEndEvent;
+import in.twizmwaz.cardinal.match.MatchState;
 import in.twizmwaz.cardinal.module.Module;
 import net.minecraft.server.v1_8_R3.PlayerConnection;
 import org.apache.commons.lang.Validate;
@@ -12,7 +16,8 @@ import org.bukkit.craftbukkit.v1_8_R3.entity.CraftPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
-import org.bukkit.event.player.PlayerMoveEvent;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerKickEvent;
 import org.bukkit.event.player.PlayerQuitEvent;
 import org.bukkit.event.player.PlayerRespawnEvent;
 import org.bukkit.event.player.PlayerTeleportEvent;
@@ -23,19 +28,33 @@ public class BossBar implements Module {
 
     private final Map<Player, FakeWither> players = Maps.newHashMap();
 
+    protected static int ENTITY_DISTANCE = 32;
+
     @Override
     public void unload() {
         for (Player player : players.keySet()) {
-            players.get(player);
+            ((CraftPlayer) player).getHandle().playerConnection.sendPacket(getWither(player, "").getDestroyPacket());
+            players.remove(player);
         }
+        HandlerList.unregisterAll(this);
     }
 
     public void sendMessage(Player player, ChatMessage message, float percent) {
         Validate.isTrue(0F <= percent && percent <= 100F, "Percent must be between 0F and 100F, but was: ", percent);
         FakeWither wither = players.get(player);
-        wither.name = checkMessageLength(message.getMessage(player.getLocale()));
-        wither.health = (percent / 100f) * wither.getMaxHealth();
-        sendWither(wither, player);
+        if (GameHandler.getGameHandler().getMatch().getState() == MatchState.WAITING || GameHandler.getGameHandler().getMatch().getState() == MatchState.ENDED) {
+            BossBar.delete();
+        }
+        if (wither == null) {
+            addWither(player, message.getMessage(player.getLocale()), true);
+            handleTeleport(player, player.getLocation(), true);
+            wither = players.get(player);
+        }
+        if (wither != null) {
+            wither.name = checkMessageLength(message.getMessage(player.getLocale()));
+            wither.health = percent * 100F / wither.getMaxHealth();
+        }
+        updateWither(wither, player);
     }
 
     private String checkMessageLength(String message) {
@@ -45,75 +64,111 @@ public class BossBar implements Module {
         return message;
     }
 
-    public void handleTeleport(final Player player, final Location location) {
-        if (players.containsKey(player)) {
-            Bukkit.getScheduler().runTaskLater(Cardinal.getInstance(), new Runnable() {
-                @Override
-                public void run() {
-                    if (players.containsKey(player)) {
-                        FakeWither oldwither = getWither(player, "");
-                        float health = oldwither.health;
-                        String message = oldwither.name;
-                        ((CraftPlayer) player).getHandle().playerConnection.sendPacket(getWither(player, "").getDestroyPacket());
+    public void handleTeleport(final Player player, final Location location, final boolean visible) {
+        if (GameHandler.getGameHandler().getMatch().getState() == MatchState.CYCLING || GameHandler.getGameHandler().getMatch().getState() == MatchState.STARTING || GameHandler.getGameHandler().getMatch().getState() == MatchState.PLAYING) {
+            if (players.containsKey(player)) {
+                Bukkit.getScheduler().runTaskLater(Cardinal.getInstance(), new Runnable() {
+                    @Override
+                    public void run() {
+                        FakeWither oldWither = getWither(player, "");
+                        float health = oldWither.health;
+                        String message = oldWither.name;
+                        if (oldWither.isVisible()) {
+                            ((CraftPlayer) player).getHandle().playerConnection.sendPacket(getWither(player, "").getDestroyPacket());
+                        }
                         players.remove(player);
-                        FakeWither wither = addWither(player, location, message);
+                        FakeWither wither = addWither(player, message, visible);
                         wither.health = health;
-                        sendWither(wither, player);
+                        updateWither(wither, player);
                     }
-                }
-            }, 1L);
+                }, 1L);
+            }
         }
     }
 
-    void sendWither(FakeWither wither, Player player) {
+    public void updateWither(FakeWither wither, Player player) {
         PlayerConnection connection = ((CraftPlayer) player).getHandle().playerConnection;
-        connection.sendPacket(wither.getMetaPacket(wither.getWatcher()));
+        connection.sendPacket(wither.getMetaPacket(wither.updateDataWatcher()));
         connection.sendPacket(wither.getTeleportPacket(getWitherLocation(player)));
     }
 
-    FakeWither getWither(Player player, String message) {
+    public FakeWither getWither(Player player, String message) {
         if (players.containsKey(player)) {
             return players.get(player);
         } else {
-            return addWither(player, checkMessageLength(message));
+            return addWither(player, checkMessageLength(message), false);
         }
     }
 
-    FakeWither addWither(Player player, String message) {
-        FakeWither wither = new FakeWither(message, getWitherLocation(player));
-        ((CraftPlayer) player).getHandle().playerConnection.sendPacket(wither.getSpawnPacket());
-        players.put(player, wither);
-        return wither;
-    }
-
-    private FakeWither addWither(Player player, Location loc, String message) {
-        FakeWither wither = new FakeWither(message, getWitherLocation(player));
-        ((CraftPlayer) player).getHandle().playerConnection.sendPacket(wither.getSpawnPacket());
+    public FakeWither addWither(Player player, String message, boolean visible) {
+        FakeWither wither = new FakeWither(player, message, getWitherLocation(player));
+        wither.setVisible(visible);
+        if (visible) {
+            ((CraftPlayer) player).getHandle().playerConnection.sendPacket(wither.getSpawnPacket());
+        }
         players.put(player, wither);
         return wither;
     }
 
     private Location getWitherLocation(Player player) {
-        return player.getLocation().add(player.getEyeLocation().getDirection().multiply(100));
+        return player.getLocation().add(player.getEyeLocation().getDirection().multiply(ENTITY_DISTANCE));
+    }
+
+    private void destroy(Player player) {
+        FakeWither wither = players.get(player);
+        if (wither == null) return;
+        ((CraftPlayer) player).getHandle().playerConnection.sendPacket(wither.getDestroyPacket());
+        players.remove(player);
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
-    public void PlayerLogout(PlayerQuitEvent event) {
+    public void onPlayerKick(PlayerKickEvent event) {
         players.remove(event.getPlayer());
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
-    public void onPlayerMove(final PlayerMoveEvent event) {
-        handleTeleport(event.getPlayer(), event.getTo());
+    public void onPlayerLogout(PlayerQuitEvent event) {
+        players.remove(event.getPlayer());
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onMatchEnd(MatchEndEvent event) {
+        BossBar.delete();
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onCycleComplete(CycleCompleteEvent event) {
+        BossBar.delete();
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onPlayerTeleport(final PlayerTeleportEvent event) {
-        handleTeleport(event.getPlayer(), event.getTo());
+        if (players.get(event.getPlayer()) != null) {
+            handleTeleport(event.getPlayer(), event.getTo(), true);
+        }
     }
 
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onPlayerRespawn(final PlayerRespawnEvent event) {
-        handleTeleport(event.getPlayer(), event.getRespawnLocation().clone());
+        if (players.get(event.getPlayer()) != null) {
+            handleTeleport(event.getPlayer(), event.getPlayer().getLocation(), true);
+        }
     }
+
+    public static void sendGlobalBossBar(ChatMessage message, float percent) {
+        for (Player player : Bukkit.getOnlinePlayers()) {
+            sendBossBar(player, message, percent);
+        }
+    }
+
+    public static void sendBossBar(Player player, ChatMessage message, float percent) {
+        GameHandler.getGameHandler().getMatch().getModules().getModule(BossBar.class).sendMessage(player, message, percent);
+    }
+
+    public static void delete() {
+        for (Player player : Bukkit.getOnlinePlayers()) {
+            GameHandler.getGameHandler().getMatch().getModules().getModule(BossBar.class).destroy(player);
+        }
+    }
+
 }
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/FakeWither.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/FakeWither.java
@@ -9,20 +9,26 @@ import net.minecraft.server.v1_8_R3.PacketPlayOutSpawnEntityLiving;
 import org.bukkit.Location;
 import org.bukkit.World;
 import org.bukkit.craftbukkit.v1_8_R3.CraftWorld;
+import org.bukkit.entity.Player;
 
 import java.lang.reflect.Field;
 
 public class FakeWither {
 
     public float health = 0;
-    public String name;
-    private float maxHealth = 200;
+    private float maxHealth = 100 / 3;
+    private float minHealth = 1;
+
     private Location location;
+    public String name;
+    private EntityWither wither;
+
     private byte xvel = 0;
     private byte yvel = 0;
     private byte zvel = 0;
+
     private boolean visible = false;
-    private EntityWither wither;
+
     private int id;
 
     public FakeWither(String name, Location location, int percent) {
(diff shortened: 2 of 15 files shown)


dentmaged:
Awesome @Alfalik!

Prodigy4532:
Nice!!

iEli2tyree011:
This looks okay, but it seems that you made it so the code is very heavily based around the actual Time Remaining boss bar. I feel a system should be managed to keep track of all boss bars, but only display the one that will end the soonest. Then, multiple boss bars can be integrated into that. @twizmwazin may have other ideas, but as I said earlier, it looks pretty good, but almost seems like the Time Remaining bar is the priority, while there are several other messages that will be displayed on the boss bar that we have to keep in mind.

Pablete1234:
In OCN you can also see the wither in the distance, if the wither's health is low

Alfalik:
@iEli2tyree011 I just believe that the Time Remaining bar is the most important and the unique that should be in Cardinal. I don't think that start or cycle bars are useful, even though that they aren't that difficult to code. I'll leave the decision to @twizmwazin.

@Pablete1234 Thanks for the info! I'll see if it can be fixed

src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
```diff
@@ -75,45 +88,127 @@ FakeWither getWither(Player player, String message) {
...
+        if (visible) {
+            ((CraftPlayer) player).getHandle().playerConnection.sendPacket(wither.getSpawnPacket());
+        }
         players.put(player, wither);
         return wither;
     }
 
     private Location getWitherLocation(Player player) {
-        return player.getLocation().add(player.getEyeLocation().getDirection().multiply(100));
+        if (GameHandler.getGameHandler().getMatch().isRunning()) {
```
ryanw-se: @Alfalik What is the point of this if/else statement? You're returning the same thing for both cases, if this is the case you don't even need the statement.

src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
```diff
@@ -75,45 +88,127 @@ FakeWither getWither(Player player, String message) {
...
+        }
+        else {
+            return player.getLocation().add(player.getEyeLocation().getDirection().multiply(110));
+        }
+    }
+
+    private void destroy(Player player) {
+        if (GameHandler.getGameHandler().getMatch().isRunning()) {
+            FakeWither wither = players.get(player);
+            if (wither == null)
```
ryanw-se: @Alfalik Your return statement can probably be consolidated to one line like so, ```if (wither == null) return;```java

src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
Alfalik: @ryanw-se Oh, it was supposed to be different, I'll correct it, thanks.

src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
```diff
@@ -75,45 +88,127 @@ FakeWither getWither(Player player, String message) {
...
     @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
     public void onPlayerRespawn(final PlayerRespawnEvent event) {
-        handleTeleport(event.getPlayer(), event.getRespawnLocation().clone());
+        if (GameHandler.getGameHandler().getMatch().isRunning()) {
+            if (players.get(event.getPlayer()) != null) {
+                handleTeleport(event.getPlayer(), event.getRespawnLocation().clone(), true);
+            }
+
+        }
+        if (TimeLimit.getMatchTimeLimit() == 0) {
```
ryanw-se: @Alfalik Your if statement checking if the match time limit is zero and then deleting the boss bar is a bit repetitious, you should consolidate it down into one statement in the handleTeleport method.

src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/FakeWither.java
```diff
@@ -40,10 +41,6 @@ public float getMaxHealth() {
         return maxHealth;
     }
 
-    public void setMaxHealth(float max) {
-        maxHealth = max;
-    }
-
```
ryanw-se: @Alfalik What was the point of moving this method? Its purely an aesthetic decision at this point and shouldn't effect your PR in any way. Same for the movement of the variables above, which were structured that way for a reason. Public variables were on top and the private ones were on the bottom, there is no point in changing this.

ryanw-se:
I did a quick review, its decent. Personally I feel the implementation is built around the time remaining bar a tad too much and would make it more challenging to implement other bars in the future. A good implementation should have support and make it easier for future boss bars to be added.

Alfalik:
@ryanw-se Thanks for the review and taking the time for do it. Now, I'll try to get the time remaining, start and cycle bars, all working. Thanks again

twizmwazin:
I'd like to see a bar system, where a Bar object can be created, and players can have a Set<Bar>. The bossbar module would then use those bars, displaying the one with the smallest time remaining.

Pablete1234:
@twizmwazin What do you mean by "how overcast does their wither so far away", they have it about 50-60 blocks away. and yes, towards the end, you can see it in the distance

TiTo-418:
https://oc.tc/forums/topics/55a824e567d52260e60041f9

Pablete1234:
Full info about the wither: http://pastebin.com/1rBhVKBt , and it's located 50 blocks away from the player

TiTo-418:
@Pablete1234 I think the withers can't have invisibility

Pablete1234:
@TiTo-418 But it's not using an "Age" tag either :/

twizmwazin:
@Pablete1234 That isn't going to work. Besides the fact that vanilla commands are entirely useless in Cardinal, the wither needs to be far from the player. The implementation system doesn't actually create a wither, it only sends packets to simulate a wither to the client. Because the Minecraft server doesn't know there is a wither we don't have to mess with any vanilla functionality.

Pablete1234:
@twizmwazin I know it's only sending the packets to do so, but if the client recives a wither, with invul 880, NoAI 1, and 50 blocks away, the client will not see it bocs it's so small, and particles don't render at that distance

twizmwazin:
@Pablete1234 No, it won't. The command tells the server how to handle the wither. In our case, there is no such wither.

Pablete1234:
@twizmwazin Even tho the wither never exists for the server, it gives an idea of what the packets sent to the player should be, now go ahead and create a packet that sends the same wither as a vanilla server having one does.

twizmwazin:
@Pablete1234 None of that information is even slightly useful when creating the packets. Creating packets is really straightforward. I've already implemented a bossbar before, although it didn't meet my standard, so I scrapped it. The issue is finding how Overcast has the wither render on the client despite being out of the render distance.

Alfalik:
@twizmwazin The max rendering distance I've tried is 112 approx., if the distance is 113, the wither won't render. Can't understand how OCN sends his wither ~200 blocks far. I'd be happy if someone find the way to.

Alfalik:
These exact particles need to be removed when ```percent < 50%```. I don't know if remove particles from a mob is a thing.

ryanw-se:
@Alfalik You can't remove the shine effect the mob gets when it reaches low health, its not possible to get rid of it unfortunately, if you go into a game on the Overcast Network you can see it in your cross hairs as well, (they didn't get rid of it either).

ryanw-se:
@Alfalik As for the max wither distance, I'm assuming it has something to do with server render distance and the client one, at 112 you're server render distance and possibly your client one would have to be set to ``7```.

Pablete1234:
@twizmwazin Again: overcast does not have the wither so far away, they have it 50 blocks away from the player, the Invul tag makes it smaller, find the way to include that in a packet

Pablete1234:
Anyone that has a tm server, do the test: 
 - Get in gamemode 1, and fly (so you don't fall)
 - /tl set 1h (to get a bar)
 - /minecraft:tp ~ ~45 ~
 - Do not move the camera, and you will see the wither's particles

CaptainElliott:
@Pablete1234 twiz has tried to explain this to you, that's not how packets works, so stop saying "find a way to include that in a packet"

Pablete1234:
@CaptainElliott But everyone is trying to find out how far they can set the wither, it's not how it works on ocn!

Alfalik:
Okay. I've made it more "globally". The thing that I need for the formula for start/cycle boss bar and can't find the way to, is the joinedTime. For example, a player do ```/start 25```, the ```joinedTime``` will be 25, and the ```time``` 25. After 10 seconds, the ```joinedTime``` will be 25 and the ```time``` will be 15. With the joinedTime I'll do the formula, that seems like this: ```float percent = (int) (time * 100F) / joinedTime```. That's what I need for start and cycle boss bars.

CaptainElliott:
@Pablete1234 how can you be so sure when you have no experience  with bukkit and it's packets?

Pablete1234:
@CapitainElliott I have 0 experience with bukkit, but go to ocn and do the tp thing, and you will see i'm not wrong about the 50 blocks thing, and I know for sure, that the client is rendering a wither with Invul tag

dentmaged:
@Pablete1234 There is no wither for the tag to be applied. It's an "illusion". Only the player can see the wither because it's sent using particles.

Pablete1234:
@dentmaged well, the "illusion" needs to include the tag, ocn does it, so it's not impossibe

dentmaged:
@Pablete1234 And how do you know OCN does it?

Pablete1234:
@dentmaged because the wither is upsidedown, that's what high Invul does

Alfalik:
Here's how OCN does his own boss bar (https://youtu.be/sxoA4ZIgba0).

Pablete1234:
@Alfalik that's exactly what i said... and everyone is saying "you can't do that with packets"

twizmwazin:
Ok, I see what it is now. Although I still don't know what "invul" means, which will be necessary to make the packet. There also needs to be a way to have it show regardless of render distance.

@ryanw-se The server's "render" distance is a server-side setting to tell the server what packets should be sent to the client. Once received by the client, it is up to the client to handle the packets and render the world.

Pablete1234:
@twizmwazin If you spawn a wither in survival, the time it takes to explode, is saved in the Invul tag (it's ticking down), when you just spawn the wither, it's small, and becomes bigger as the invul ticks down, when it reaches 0, the wither explodes and "the battle starts" (if you are in survival), when setting this value higher than it usually is, it makes the wither smaller and smaller, to the point where it's invisible (880) and even upside down (what ocn uses: 1000)

Alfalik:
@twizmwazin This is what invul does (https://youtu.be/h1mqy1BaitU), but sportbukkit don't have anything related to invul. Here's the command if needed ```/summon WitherBoss x y z {NoAI:1,Silent:1,Invul:1000,Health:80F}```.

Pablete1234:
@Alfalik noo!! what did you do!! vanilla?? never use vanilla!?!?!?!? you may say "But now I know how invul works and affects the wither" well:
"No, it won't. The command tells the server how to handle the wither. In our case, there is no such wither. Also, you need to take into account render distance."

twizmwazin:
@Pablete1234 Really? Be mature. The vanilla command doesn't help me much at all, however an understanding of this "invul" tag does. Unless you are the one actually writing this and know what you are doing, I'd advise minding your own buisness.

Pablete1234:
@twizmwazin maybe i'm being immature, but I've been talking about that for 2 days and everyone has ignored me instead of trying to run my tests, @Alfalik just recorded what i had said, and now you understeand, but you didn't care at all of what I said

twizmwazin:
Even if the it is the same information, Alfalik presented it in a way I could understand. I really couldn't figure out what you were trying to say other than that you could use a vanilla command to summon a wither.

Aaron1011:
The `Invul` tag data is sent to the client through entity metadata: http://wiki.vg/Entities#Wither

Alfalik:
@Aaron1011 Thanks! That's what I needed

twizmwazin:
An implementation written by someone else that seems to work fairly well from my brief testing. https://bitbucket.org/inventivetalent/bossbarapi/src/407c3fc2e5c887065c35596d9e382c895ed3d188/BossBarAPI/src/org/inventivetalent/bossbar/BossBar.java?at=master

twizmwazin:
I have a handful of things I want to do for bossbars, I'm going to try and get to them, this may need to work around them.

Alfalik:
@twizmwazin I've got to work start, cycle and time remaining boss bars. But, a NPE is thrown when they are sent. Also, need to find the right formula to get percent on start and cycle boss bars. Hopefully you can test the changes.

Pablete1234:
@Alfalik when someone starts a countdown, save the total time, and then percentatge = time remaining / total time * 100

Alfalik:
@Pablete1234 I know the formula, the unique problem is how to get the joined time, or as you said, total time.

dentmaged:
@Alfalik https://github.com/Alfalik/CardinalPGM/blob/2c3b64127bfbe8a4dbf49cbdeeec1725a5c48b40/src/main/java/in/twizmwaz/cardinal/module/modules/cycleTimer/CycleTimerModule.java#L23

Just create another integer named originalTime then add ```setOriginalTime(time);``` in this method: https://github.com/Alfalik/CardinalPGM/blob/2c3b64127bfbe8a4dbf49cbdeeec1725a5c48b40/src/main/java/in/twizmwaz/cardinal/module/modules/cycleTimer/CycleTimerModule.java#L102

Alfalik:
@dentmaged Thanks, it works.

The unique and actual problem is a NullPointerException thrown when a boss bar is created. I don't know what's wrong with the code or wheres the error. Hope anyone can help me.

dentmaged:
@Alfalik You're welcome. I encountered the same issue whilst working on Cardinal+. I can't remember how I fixed it. Just use ```System.out.println(something == null);```

Alfalik:
A quick update:

+ The NPE were fixed by @dentmaged.
+ Add a description after every metadata of the wither.
+ Remove isState and use getState() == State instead.
+ Int originalTime is everytime 30, I'll fix it as soon as possible.

Pablete1234:
@Alfalik the wither needs to go closer to the player, playing at 5 chunks it flashes, and less than 4 chunks it won't render at all.

twizmwazin:
Did some brief testing, I'm pretty impressed with this so far. Besides Pablete's issue, the only other issue I noticed was that you assume that the start countdown is 30. You will need to make the max whatever the user entered as the start time.

Alfalik:
As @Pablete1234, the wither distance were changed to 50 blocks from the players. About the match start message, it is a half of a second, I don't see other way to do it or the problem itself with it.
@twizmwazin Still trying to fix ```originalTime``` on start timer being 30.

twizmwazin:
You probably want to put the wither around 20 blocks away, so it can be seen from the smallest render distance.

Alfalik:
@twizmwazin Done. I'll be working on the start issue tomorrow.

Pablete1234:
@Alfalik the 1 second remaining message should stay for 1 second, and the match starting now should be the "0 seconds remaining" message (if there was one) and should stay until 1 second of the match has passed.
ps: In OCN they removed the "match starting now", it's just "match starting in 0 seconds" , and it doesn't even show up if it's a match with time limit, it's like lowest priority

Alfalik:
@Pablete1234 I just removed it. Thanks for meticulous testing although.

Alfalik:
Yess

ryanw-se:
@Alfalik Nicely done! :clap:

Pablete1234:
@Alfalik :clap:

Alfalik:
@Pablete1234 The ```show-before="time"``` is only for monument modes, and it will show the one with less time.
About the click sounds, could you find the exact sound, volume and pitch of it? That'll be helpful.

TiTo-418:
@Alfalik Look at https://github.com/twizmwazin/CardinalPGM/issues/227

Alfalik:
Update:

+ Got to work the monuments modes bar, but it still have some errors:
    - It doesn't work without time limit.
    - Need to reset time of timeBeforeMode after every monument mode runs. That's why (if anyone test) only the first mode is shown in the boss bar.
+ On start timer, ```originalTime``` always return 30 seconds.

If somebody could help with any of these inconsistencies, it would help much.

@TiTo-418 It isn't there.

iEli2tyree011:
@Alfalik Monument modes always have a name

dentmaged:
@Alfalik When monuments have no name define the XML it is the blocks in uppercase without underscores in the colour red.

Alfalik:
@iEli2tyree011 I'm not that sure, look at docs https://docs.oc.tc/modules/monument_modes
@dentmaged Thanks

iEli2tyree011:
@Alfalik 
> Monument modes always have a name

Alfalik:
Well, the monument modes boss bar totally works, except if the match does not have a time limit. I still can't understand the start timer issue that return always 30 seconds. The rest is working perfectly fine.

AngelCasillas01:
Maybe this works in cannon duel too

TiTo-418:
@AngelCasillas01 
> Well, the monument modes boss bar totally works, *except if the match does not have a time limit*.

AngelCasillas01:
@TiTo-418 Comfirmando :p

Prodigy4532:
@AngelCasillas01 "Confirmando"*

Alfalik:
A quick update. I've been making the boss bar more consistent (solving bar health flashing and the wither being visible sometimes), things that have been totally fixed. I still have two issues that really not have reviewed:

+ Int ```originalTime``` on Start Timer always return to 30 seconds.
+ Monument modes bar not working if the match doesn't have a time limit.

Hope anyone can help or tell me some issues (if found).

TiTo-418:
@Alfalik Can you, with the boss bar active, do /cycle 0? I think you will find an issue, i can not describe it

Alfalik:
@TiTo-418 Could not find the issue. Can you explain it?

TiTo-418:
@Alfalik I saw the wither particles, it is very fast. Im not sure, maybe im crazy. I'll try to record it.

Alfalik:
@TiTo-418 Yes. Those particles are sent when the wither die.

TiTo-418:
@Alfalik The boss bar disappear if you die and respawn

Alfalik:
@TiTo-418 Could you ```git clone``` again and test, I've ammend the commit

TiTo-418:
@Alfalik Sure

TiTo-418:
@Alfalik Still happen

twizmwazin:
I'm taking a look at this, couple of things:
 - Use proper package naming conventions. It is standard to use camelCase, you have a package inproperly named `Reflection`. Regardless this would be more suited in util.
 - Wildcard imports. I don't like them usually, and importing the whole Minecraft server is a prime example of why not. It may have been an accidental thing done automatically by your IDE, but it should be configurable, I know for sure it is in IntelliJ.
 - Back on reflection, I don't really want to use reflection when it isn't necessary. It is ok when accessing otherwise private fields and methods, although we don't need it for NMS classes. We can just access them directly, it is already done by the plugin.

Alfalik:
Thanks for taking a look, and about the things that you pointed:

+ I never change/touch imports, IntelliJ replace them itself with wildcards, automatically, as you said. I've disabled this function altough.
+ Should I move reflection to util and leave it for future things, or directly remove it? I find it useful and easy to understand.

twizmwazin:
I'd just remove the reflection classes altogether.

Alfalik:
@twizmwazin Done
@TiTo-418 The wither disappearing issue if a player dies have been totally fixed.

Alfalik:
I've finished implementing all boss bars (start, cycle, time remaining and monument modes). The two issues that I had, are totally fixed. I'd be happy if anyone can further test and tell me if it is completely ready to be merged or need some enhancements.

TiTo-418:
@Alfalik The particles problem is not fixed

TiTo-418:
<a href="http://imgur.com/r3xPj7Q"></a>

TiTo-418:
Sometimes it flashes

Alfalik:
@TiTo-418 About your points:

+ It isn't a problem.
+ I'd recommend to test with a server, such as hosthorde, with a minium of 2 GB of RAM, not a localhost. Also define how it flashes for you, the bar return to the lower health or it dissapear totally.
+ It appears for me.
+ And for the image, I'll take a look, I misspelled something.

Alfalik:
@TiTo-418 Could you test again?

TiTo-418:
@Alfalik It works perfect now :D

Alfalik:
@twizmwazin Any thoughts?

twizmwazin:
I can do some testing in a bit, do you think this is ready to be merged?

TiTo-418:
@twizmwazin Yes, it was tested today and works fine

Alfalik:
@twizmwazin Yes, I've fixed all the errors or inconsistencies.

Pablete1234:
@Alfalik OCN uses one wither and keeps changing the name, health and position of that one, instead of summoning a new one each time

Alfalik:
@Pablete1234 That's how this boss bar works. The wither is only sent on this events https://github.com/Alfalik/CardinalPGM/blob/bossbar/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java#L144, and the wither is updating constantly.

Pablete1234:
If you /ready and then /unready the wither stays there

Alfalik:
Try now

twizmwazin:
I'm under the impression that this is ready to be merged, @Alfalik can you rebase this?

Alfalik:
@Pablete1234 The monuments boss bar staying until the match end issue, is now fixed.
@twizmwazin Done!

Pablete1234:
The commit message deserves to be "Implement a boss bar system" it's not basic anymore

Pablete1234:
I want to point out that ocn does not have /tl set 1h, ocn only has /tl 1h, and means in the moment you type the command, the countdown starts going, that means after a /tl command the bossbar always gets to full health
In cardinal, you can /tl set 1h, when 50 mins have passed, and the bossbar won't start at  max health, ti will start at 10/60 mins

I don't know what's the prefered behaviour, but i'll let you know guys

dentmaged:
@Pablete1234 When @twizmwazin (or @iEli2tyree011) was writing the ```/timelimit``` command, he asked if it should be ```/timelimit <time>``` or ```/timelimit <set|add>```. I prefer ```/timelimit <time>```, I might rewrite it in a PR (if I get enough time).

Pablete1234:
@dentmaged but with /tl set you can choose exactly how long the match is going to be, for example, if a match starts with no tl, and at the mark of 50 mins you realize you want to set the tl, in ocn you can't ensure the match will end at the 1h mark exactly, in cardinal you can
Maybe the best behaviour would be that /tl set 60m on 51:24 is an "alias" of /tl 8m36s, and /tl add 10m (just after the first one) is an "alias" of /tl 18m36s

dentmaged:
@Pablete1234 The ```/timelimit``` command is buggy and needs to be rewritten. The time remaining messages are misfiring after setting the time limit.

Pablete1234:
@dentmaged i know, but that's more about the time notifications, they fire when you /tl, and they schedule themselves to fire again 5 mins later, instead of firing on the command, and on the next 5 multiple time

twizmwazin:
merged

PR #770, opened 2015-07-17
Add shortest commands for team commands

Alfalik:
I've added shortest "aliases" for team commands. Also, now the shuffle command can only be executed if the match is waiting or starting and the team change name is sent to the global channel.

Here's a test https://youtu.be/c0Q9Q1gs068


dentmaged:
@Alfalik You don't need an extra class, just register TeamCommands and TeamCommands.ParentTeamCommand

Alfalik:
@dentmaged Thanks! Didn't really know.

Alfalik:
@twizmwazin Could you add this? For me, it's pretty useful do ```/force <player> <team>``` instead of ```/team force <player> <team>```. I also fixed shuffling teams during a match being played.

twizmwazin:
@Alfalik Can you rebase the commit on master? After that I looks ready to merge

Alfalik:
@twizmwazin Done!

Issue #566, opened 2015-04-09
Staff isn't whitelisted automatically


TiTo-418:

dentmaged:
Cardinal uses Bukkit's built-in whitelister, not a custom one.

CaptainElliott:
we do use a custom whitelister https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/command/WhitelistCommands.java

most of it was made by OCN devs but it has been editied for our needs and I plan to make some additions to it also using @ShinyDialga45 unofficial OCN api. I will also implement this

dentmaged:
@CaptainElliott really? O_o

Why does https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/command/WhitelistCommands.java#L52 refer to ```Bukkit.getServer().reloadWhitelist()```? Can you set a custom whitelist object in Bukkit?

CaptainElliott:
@dentmaged I'm sorry, I thought you just meant using the default white list stuff.
But there is really no other way to do a whitelist, especially since the api bukkit has is so flexible

Eudaldca:
If this was done, could it be made via a permission node? This way it can also be applied for other ranks, or removed if you don't want it.

dentmaged:
@Eudaldca I think it may use ```rank.isStaffRank();```

PR #771, opened 2015-07-18
Add ranks system

iEli2tyree011:
Adds a rank system to Cardinal. Still working on finishing touches and adjustments.


TiTo-418:
YEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEES :DDDDDDDDDDDDD

Alfalik:
From testing:

+ Permissions nodes aren't working for me, could you confirm?
+ Do you plan on adding inheritance?
+ A default rank should be added with default permissions, like ```
<permission>cardinal.chat.global</permission>```

Pablete1234:
yay \ö/

TiTo-418:
Cool :D
<a href="http://imgur.com/861X3wx"></a>

iEli2tyree011:
@Alfalik Permissions are working for me, and inheritance is planned to be added. A default rank will be added as well.

TiTo-418:
I think a /ranks list is needed :P

TiTo-418:
I can not start with this xml: http://pastebin.com/psJH2juY
I dont know if it has errors :P

iEli2tyree011:
@TiTo-418 What do you mean? Are there errors in the Console, or does it just not list the ranks under /ranks list? I can't fix a problem if I don't know what's wrong :stuck_out_tongue:

TiTo-418:
The server does not start it froze on "Deleting match files..."

twizmwazin:
@TiTo-418 That shouldn't be affected by this PR. If deleting old match files takes too long for you, it can be disabled in the config.

TiTo-418:
@twizmwazin With an empty ranks.xml the server start normally

iEli2tyree011:
@twizmwazin That's just the last message that is logged before the server freezes, the problem is with inheritance logic.

TiTo-418:
I don't know if this is intended, but i can not give or remove ranks from offline players http://prntscr.com/7u1mc5

iEli2tyree011:
@TiTo-418 The freezing issue has been solved, your ranks should work.

TiTo-418:
@iEli2tyree011 Yes, it is fixed, thanks!

TiTo-418:
Ok, i don't know if i removed default perms, but streamers can talk in global chat (same ranks.xml)

iEli2tyree011:
@TiTo-418 Just because the `Streamer` rank's parent isn't `Default` doesn't mean the player doesn't have the `Default` rank, which gives the player chat permissions. Since the `Default` rank is default, the rank is given to any player when they join the game.

TiTo-418:
@iEli2tyree011 Oh, ok. There are a way to remove it?
And, is necesary the Default parent in the Moderator Rank?

iEli2tyree011:
@TiTo-418 Not that I can think of. And since the `Default` rank is given to all players when they join regardless of having the `Moderator` rank or not, inheritance in that case shouldn't be necessary.

TiTo-418:
@iEli2tyree011 Ok, thanks

iEli2tyree011:
@TiTo-418 Okay, thank you :smile:

dentmaged:
@iEli2tyree011 Can players revoke permissions by add a ```-``` in front of a permission? I must say, this looks awesome!

iEli2tyree011:
@dentmaged That's actually a really good way of condensing that. I'm thinking that would override any permissions received from inheritance, which would solve @TiTo-418's problem

iEli2tyree011:
I added it so you can disable permissions for ranks, which overrides permissions gotten from inheritance and is not inherited itself. This is untested, and may not work. This is just what I have so far for the night, but it's getting late, and I'm going to go. If someone could get back to me on how this works, that would be great.

TiTo-418:
The disabled permissions works fine.
The command /rank info isn't implemented

Alfalik:
Could you fix #566 with a toggleable option in the ```config.yml```, like ```whitelist-staff = true```

TiTo-418:
@Alfalik That souldn't be a toggleable option

Pablete1234:
It would be nice if you could add a bypass-whitelist permission, or something like that (just the same way op's can bypass it)

TiTo-418:
@Pablete1234 https://github.com/twizmwazin/CardinalPGM/issues/566

TiTo-418:
I think the ranks branch can be removed

PR #783, opened 2015-07-20
Fix force cyling to a map, fixes #781

### LINKED ISSUE #781: Forcing cycle with map fails
[This](http://youtu.be/fl5LZ5xjWY4).
Version:
screenshot_223

CaptainElliott:
This is the only fix for https://github.com/twizmwazin/CardinalPGM/issues/781 I could think off.

The issue is that getting the joined string also contained the flag `-f`

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/command/CycleCommand.java
+++ b/src/main/java/in/twizmwaz/cardinal/command/CycleCommand.java
@@ -12,6 +12,7 @@ import in.twizmwaz.cardinal.module.modules.cycleTimer.CycleTimerModule;
 import in.twizmwaz.cardinal.module.modules.timeLimit.TimeLimit;
 import in.twizmwaz.cardinal.rotation.LoadedMap;
 import in.twizmwaz.cardinal.util.ChatUtil;
+import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.command.CommandSender;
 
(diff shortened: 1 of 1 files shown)


twizmwazin:
Perhaps rather than replacing it directly, can we check if the last argument is a flag and then replace it then? Theoretically a map name could be "Ipsum-florem"

CaptainElliott:
@twizmwazin the issue then is that a flag can be at any point in the command, not necessarily last

CaptainElliott:
what if we checked it for being standalone so 
` -f ` with the extra white space, that way we know for certain it's a flag

Pablete1234:
Maybe search for a " -f  " space before and after the flag, or " -f"+end (without space, but at the end of the command)

twizmwazin:
The flag should be respected as such outside of the map name. If someone tries `/cycle 5 Race for -f Victory 2`, it should say map not found. However `/cycle 5 -f Race for Victory 2` or `/cycle 5 Race for  Victory 2 -f` should both work.

Pablete1234:
Then just try to find it after the number, (3rd argument) or at the end

CaptainElliott:
@twizmwazin the issue only arrises when we used `#getJoinedString()` if we get say the string at index 0 from this command `/cycle -f 0` it will return 0, so what if instead of using `getJoinedString` we just keep appending the map name upntil there are no other indexes left or it reaches a `-f` on it's own

twizmwazin:
if you can address those two line comments I'll merge this

PR #794, opened 2015-07-24
Fix wl team Command, fixes #710

### LINKED ISSUE #710: whitelist team doesn't work properly
/wl team <team name> does the same as /wl all, it doesn't properly test for the right team

to reproduce:
join blue team, make sure you are not on the whitelist already
type /wl team red
you will be added to the whitelist while being on blue

tested in: 4f2c109

Pablete1234:
Fixes #710 , and allows for "/wl team the red team" (before you could only add 1 argument)

Tested code


iEli2tyree011:
Can you amend the commit and perhaps make the commit message "Fix wl team command"? It's better to not have a `/` in a commit message, just for convention's sake.

Pablete1234:
@iEli2tyree011 done

twizmwazin:
merged

PR #806, opened 2015-07-29
Fix kit health, food and saturation

Pablete1234:
Health : 1-20
Food: 0-20
Health is no longer the max health, it's the player health
Health, food and saturation don't get added, they get set if they are bigger than the actual player's health/food/saturation, unless forced
Taken from: https://docs.oc.tc/modules/kits

Test video: http://youtu.be/caXbso5aOAQ
Xml Used : http://pastebin.com/zmYHrXPr
The slot of the sword tells what region i'm in

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/kit/Kit.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/kit/Kit.java
@@ -77,18 +77,18 @@ public class Kit implements Module {
         } catch (NullPointerException e) {
 
         }
-        if (health > -1) {
-            if (force) player.setMaxHealth(health);
-            else player.setMaxHealth(player.getMaxHealth() + health);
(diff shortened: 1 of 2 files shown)


SungMatt:
Won't this break maps like Dust where there are only 2 hearts displayed rather than 10 hearts?

Pablete1234:
@SungMatt maps don't use the health tag to lower the max health, they use negative health boost

twizmwazin:
can someone get results with dust?

Pablete1234:
@twizmwazin i have a map that uses same effect as dust and works as expected

twizmwazin:
So do we think this is good to merge? @iEli2tyree011

Pablete1234:
@iEli2tyree011 fell out of the world

twizmwazin:
merged

PR #837, opened 2015-08-17
Add /me command, closes #836

### LINKED ISSUE #836: /Me and /mE can be used while the global mute is enabled

Alfalik:
It should be better to simply create and override the ```/me```, instead of blocking it. Here's an image of how it looks like https://gyazo.com/dc6c7193eeb418ab816eaedd1508a9b3


iEli2tyree011:
You can still use `/minecraft:me` no matter what. We can't overwrite that command.

Alfalik:
@iEli2tyree011 That's why I added this https://github.com/twizmwazin/CardinalPGM/pull/837/files#diff-c2eee098f5e1d9f71d4efd97974e23a2R5

iEli2tyree011:
@Alfalik Did you test that? I've always had problems removing default permissions with the permission module.

Alfalik:
@iEli2tyree011 Meh, players can still use default minecraft permissions, ```minecraft:me``` in this case.

twizmwazin:
I'll be honest I'd rather just outright disable the vanilla command

Pablete1234:
what about just a ```#toLowerCase.startsWith(/me)```

TiTo-418:
This PR can be closed

Pablete1234:
it still would be nice to add a cardinal /me so that it uses display names (with ranks and team) instead of white names, but the vanilla one should be left as it is (usable by default, and blocked during mute or gmute) and the same behaviour should be used for cardinal:me

Alfalik:
I've ammend the commit, so if a player do ```/minecraft:me```, the player won't have access. Also, a ```/me``` command has been added (https://gyazo.com/c584771128c1a7ed04d39871b2d5cb97). As @Pablete1234 said, it display the actual rank and team of the player.

Pablete1234:
Leave the old chat module as it was, if someone has their server on the bukkit setting that uses vanilla by default it will be better to also block the regular /me (bcos it will be the vanilla one)

Alfalik:
The vanilla command is now totally disabled, and I've added a ```/me``` command that shows the rank and team of the player.

unixfox:
:+1:

src/main/java/in/twizmwaz/cardinal/command/BroadcastCommands.java
```diff
@@ -23,4 +27,13 @@ public static void say(final CommandContext cmd, CommandSender sender) throws Co
     public static void broadcast(final CommandContext cmd, CommandSender sender) throws CommandException {
         ChatUtil.getGlobalChannel().sendMessage(ChatColor.RED + "[Broadcast] " + cmd.getJoinedStrings(0));
     }
+
+    @Command(aliases = {"me"}, desc = "Send a global message starting with your name", min = 1, usage = "<message>")
+    @CommandPermissions("cardinal.me")
+    public static void me(final CommandContext cmd, CommandSender sender) throws CommandException {
+        if (!(sender instanceof Player)) throw new CommandException(new LocalizedChatMessage(ChatConstant.ERROR_CONSOLE_NO_USE).getMessage(ChatUtil.getLocale(sender)));
+        if (!GameHandler.getGameHandler().getGlobalMute()) {
```
Pablete1234: ```if (!GameHandler.getGameHandler().getGlobalMute() || sender.hasPermission("cardinal.globalmute.override")) {```

SungMatt:
Should have been fixed in 6ce8cbd , thanks for your work in this PR!

Pablete1234:
@SungMatt it isn't quite fixed, but i agree thar 1.9 fixed quite alot of stuff with player selectors (try /minecraft:say @a[team=observers]) now bukkit properly uses scoreboard team color / names or something, but it's still not the bukkit display name, it's just the scoreboard team color

PR #845, opened 2015-08-24
Prevent opening crafting tables before the match starts

### LINKED ISSUE #788: Observers can use crafting tables

Pablete1234:
If you join a team but the match hasn't started or has ended, you can no longer open crafting tables
fixes #788


twizmwazin:
merged

PR #847, opened 2015-08-25
Allow restricted kits to be used by ops only, fixes #846

### LINKED ISSUE #846: Problem with Classes "restrict" tag
Docs.oc.tc says:
"restrict : If set to false only OPs can use the class."
but https://maps.oc.tc/DTM/Standard/DATAbasin/map.xml, has "restrict" set to false, and players can still get the kits
I'm guessing maybe they changed it in the proto?? (databasin is proto 1.3.6 and docs is proto 1.4)

If it's about proto, cardinal should probably try to support both 1.3.x and 1.4 proto, since alot of maps still use proto 1.3

dentmaged:
This hopefully fixes #846


twizmwazin:
merged, thanks

PR #848, opened 2015-08-25
Update TeamPicker

### LINKED ISSUE #829: Team picker not updated
This has recently been added to PGM, when you join a team before the match starts it will no longer remove the helmet picker. Only when you have joined, it will show the boots in the team picker.

Pablete1234:
Update TeamPicker to add the leave boots
fixes #829


twizmwazin:
merged, thanks

PR #849, opened 2015-08-25
Prevent observers stopping the countdown if observers-ready is true

### LINKED ISSUE #711: Observers can ready even when the setting is disabled
i have the setting:
```observers-ready: false```
if you join red and type /ready, it says "red is ready"
if you join blue and type /ready, it says "blue is ready"
if you join obs and type /unready, it says "your team is already already not ready"

the thing is you can just type /ready on obs, observers will become ready, and then you can /unready, and the start will be canceled

IMO what should happen is that if you have the setting set to false, observers should be ready by default (so if you type /ready it says "your team is already ready"), and it should not allow players on observers to do /unready

Pablete1234:
If observers-ready is set to false, observer can't /ready (tram already ready) and can't /unready (team can't unready)

Fixes #711


twizmwazin:
merged

PR #850, opened 2015-08-25
Prevent pistons bypassing max building height

Pablete1234:
Pressing that button would before push the piston and the slime blocks into max build height, with this PR both get cancelled

Tested, ready to merge


twizmwazin:
merged

PR #859, opened 2015-08-29
Fix wl team command

Pablete1234:


src/main/java/in/twizmwaz/cardinal/command/WhitelistCommands.java
```diff
@@ -151,21 +151,17 @@ public static void kick(final CommandContext args, final CommandSender sender) t
...
-        }
-        msg = msg.trim();
+        String msg = args.getJoinedStrings(0).toLowerCase();
         if (Teams.getTeamByName(msg) == null) {
             throw new CommandException(new LocalizedChatMessage(ChatConstant.ERROR_NO_TEAM_MATCH).getMessage(ChatUtil.getLocale(sender)));
         }
         for (Player player : Bukkit.getOnlinePlayers()) {
             Optional<TeamModule> team = Teams.getTeamByPlayer(player);
-            if (team.isPresent() && team.get().getName().startsWith(msg)) {
+            if (team.isPresent() && team.get().getName().toLowerCase().startsWith(msg)) {
```
iEli2tyree011: This is still broken, we should be comparing the teams, not the teams' names

Pablete1234:
@iEli2tyree011 what if you have "Team red" and "Team blue", and you do /wl team team, both will get whitelisted, as both start with "team", i think this is intended behaviour

TiTo-418:
https://github.com/twizmwazin/CardinalPGM/pull/794

Pablete1234:
@twizmwazin opinions? i think this is a good enough fix, it will wl anyone that's on a team that starts with what you type

src/main/java/in/twizmwaz/cardinal/command/WhitelistCommands.java
```diff
@@ -151,21 +151,17 @@ public static void kick(final CommandContext args, final CommandSender sender) t
         ChatUtil.getGlobalChannel().sendLocalizedMessage(new UnlocalizedChatMessage(ChatColor.RED + "{0}", ChatConstant.GENERIC_KICKED_NOT_WHITELISTED.asMessage()));
     }
 
-    @Command(aliases = {"team"}, desc = "Adds everyone on a team to the whitelist.", max = 1, min = 1)
+    @Command(aliases = {"team"}, desc = "Adds everyone on a team to the whitelist.", min = 1)
     @CommandPermissions("cardinal.whitelist.team")
     public static void team(final CommandContext args, final CommandSender sender) throws CommandException {
         int count = 0;
-        String msg = "";
-        for (int i = 2; i < args.argsLength(); i++) {
```
dentmaged: ```msg = args.JoinedStrings(0);```

Pablete1234: but sending it to lower case also helps

dentmaged: How does it "help"?

Pablete1234: Team is called "Red Team" and you type /wl team red

dentmaged: And you should get the team specified in the args with ```Teams.getTeamByName(msg);```

Pablete1234: @dentmaged as i said  b4, if there is a team called "Team Red" and a team called "Team Blue", /wl team team should wl both, not pick the first one (alphabetically) and wl them

Pablete1234:
@iEli2tyree011 @twizmwazin i've redone this, if there are 2 teams called "Team Red" and "Team Blue" and you /wl team team, it will throw an error "Multiple teams matched querry : Team Red, Team Blue."

Pablete1234:
@iEli2tyree011 @twizmwazin @CaptainElliott needs testing tag? some ppl should test this and show the results, IMO it's ready to merge

TiTo-418:
All works fine

twizmwazin:
I see some formatting that needs to be fixed.  Also, I don't think we should be removing spaces when comparing team names to the user input.

Pablete1234:
@twizmwazin fixed 2 missing spaces (in formatting), and added the "replace(" ", "")" to the team name (missing part).
and yes, i think we should compare teams without spaces, because it makes the system more user-friendly, the same way you remove capital letters so that you don't have to writte it exactly as it is in the team name, i think getting rid of spaces also helps in that

twizmwazin:
Rethinking over it you are probably right on the spaces issue.

twizmwazin:
merged

PR #863, opened 2015-09-01
Add some enhancements to monument modes boss bar

Alfalik:
This should fix be the bar being deleted and sent at the same time, when all modes has ran. I'd like anyone to confirm testing

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/monumentModes/MonumentModes.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/monumentModes/MonumentModes.java
@@ -103,11 +103,11 @@ public class MonumentModes implements TaskedModule {
                     showBefore = mode.getShowBefore();
                 }
             }
-            if (timeBeforeMode < showBefore && name != null) {
+            if (timeBeforeMode > 0 && timeBeforeMode < showBefore && name != null) {
                 int percent = (int) (timeBeforeMode * 100F) / showBefore;
                 BossBar.sendGlobalBossBar(new UnlocalizedChatMessage(ChatColor.RED + "{0}", new LocalizedChatMessage(ChatConstant.UI_MODE_IN_TIME, new UnlocalizedChatMessage(ChatColor.RED + name + ChatColor.AQUA)), ((timeBeforeMode) == 1 ? new LocalizedChatMessage(ChatConstant.UI_SECOND, ChatColor.DARK_AQUA + "1" + ChatColor.AQUA) : new LocalizedChatMessage(ChatConstant.UI_SECONDS, ChatColor.DARK_AQUA + String.valueOf(timeBeforeMode) + ChatColor.AQUA))), percent);
             }
-            if (timeBeforeMode <= 0 || (TimeLimit.getMatchTimeLimit() == 0 && timeBeforeMode > showBefore) || name == null) {
+            if (timeBeforeMode < 0 || (TimeLimit.getMatchTimeLimit() == 0 && (timeBeforeMode > showBefore || timeBeforeMode <= 0))) {
                 BossBar.delete();
             }
         }


unixfox:
:+1:

twizmwazin:
@Pablete1234 If you could confirm if this fixes your issue, I'll merge this.

Pablete1234:
@twizmwazin it works, but I think the wither distance from the player should be bumped up to prevent the player seeing the particles

unixfox:
@Pablete1234 yes like that : http://i.unixfox.eu/2015-09-01_15-07-05.gif

Pablete1234:
@unixfox like that, and also while falling from high places

twizmwazin:
@Pablete1234 It needs to be kept at a certain distance so users even with 2 chunks render distance can see the boss bar.

Pablete1234:
@twizmwazin maybe the server should check the render distance of the client to see if it's really small, and if so leave it like that, or else move it away 50 blocks
idk if it's possible, but i don't think the server sends more chunks than what the client requests, so i think it should be possible

iEli2tyree011:
@twizmwazin I believe Overcast keeps it at 32 blocks away (2 chunks) but I'm not sure if there's more problems with not seeing the boss bar when you make your render distance that low.

Pablete1234:
@iEli2tyree011 they have it 50 blocks away (when we had tm i did the test), look straight up, /minecraft:tp ~ ~45 ~ and you could see the particles

dentmaged:
@Pablete1234 It's probably at 48.

Alfalik:
I don't think this is related with the actual issue, but well. Should I change it to 48 blocks far or keep it as it's now (32)?

twizmwazin:
@Alfalik Unless you can find a reliable way to get client render distance, keep it at 32 blocks.

Pablete1234:
@Alfalik how about the pos of the wither instead of being (pos of the player + vector of where he is looking*32) be (pos of the player + vector of where he is looking*32 + vector of movement of the player)
that would mean, if the player is moving down, the wither would spawn lower, if player is moving foward, the wither will spawn foward

Alfalik:
I won't have the time to code or use a computer until 17 of september (I'm going to the United States). Hope you can code a thing that get the render distance of a player @Pablete1234.

Pablete1234:
@Alfalik this PR fixes monument modes, but not core modes

Alfalik:
Aren't they the same?

Pablete1234:
@Alfalik they should be, but it still happens on monument modes with this PR

Pablete1234:
@twizmwazin @Alfalik when shooting you sometimes see the arrow fall down (because it's hitting the wither client-side) and it is a very annoying bug, what could fix it? *cough* 48blocks *cough* either that or moving the player a bit up/down so it's not on the sight of arrows

Alfalik:
@Pablete1234 Create a new issue. I don't think this is related with the actual issue, besides, I won't fix the arrows hiting the fake wither or you seeing the particles, because I don't have the time to. Thanks
Back to the topic, does this pull request fix the actual issue?

Pablete1234:
@Alfalik that was on a server that i had putted a cardinal ver with this fix, but someone else "updated" cardinal to a version without it
This PR does fix the wither blinking issue, but it's called "enhancements" so yeah, another enhancment won't hurt anyone

Alfalik:
@Pablete1234 In-case you will fix the wither particles being rarely and arrows hiting the wither, use this http://wiki.vg/Protocol#Client_Settings

Pablete1234:
@Alfalik the other common issue in bossbars atm is that long tl bossbars don't get shown the last 5 seconds

Alfalik:
It's a math problem. This is the actual formula:

```java
int percent = (int) ((100F * timeLeft) / TimeLimit.getMatchTimeLimit());
```

If the int ```timeLeft``` is a low number compared with a high number (in this case a long time limit as you said), it'll return to 0F, and the bar won't be show. Here's a clearly example with a 3 hours time limit:

```
int percent = 5s / 10800s = 4,69x10^-4 = 0,000469 We'll ignore the percent, we want the number
```

It's a very little number, so it will be approximate to 0.

twizmwazin:
merged

twizmwazin:
As an additional note, to fix @Pablete1234's mentioned issue about the bar dissapearing, mathematically it makes sense to approximate to 0, however the health should always be kept at 1, even if it should be less than this, to keep showing the bar. Someone will need to do this.

SungMatt:
@twizmwazin So basically if `health=0` add one?

twizmwazin:
@SungMatt Yeah, pretty much. Not really anything complicated.

PR #864, opened 2015-09-01
Send team alias rename message to global chat

Pablete1234:
Broken by the command branch
Before it would send everyone the message in the lang of the sender, now everyone recives it in their own lang


TiTo-418:
Tested and works fine

twizmwazin:
merged

PR #870, opened 2015-09-02
Use proper default fly speed

Pablete1234:
Atm when using cardinal you fly at double the speed, and it's annoying for building etc


iEli2tyree011:
Thanks

PR #872, opened 2015-09-04
Fix wither dissapearing when health is too low

Pablete1234:
I hope this fixes the issue
I didn't test this
(Maven can't build because of a dependency issue)

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/bossBar/BossBar.java
@@ -52,7 +52,7 @@ public class BossBar implements Module {
         }
         if (wither != null) {
             wither.name = checkMessageLength(message.getMessage(player.getLocale()));
-            wither.health = percent * 100F / wither.getMaxHealth();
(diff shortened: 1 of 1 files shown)


twizmwazin:
You should be able to build now, test it when you can.

Pablete1234:
@twizmwazin i'll be away from home 3 days, if someone else can test this it would be nice

dentmaged:
@twizmwazin Can you restart the build? If not, @Pablete1234 can just ```git commit -m "Fix wither disappearing when health is too low" --amend```, making sure there are no new additions.

Pablete1234:
@dentmaged I did what you said, but there is still some dependency that doesn't work
@twizmwazin tell me when cardinal should build again, and i'll re-amend it

dentmaged:
@Pablete1234 oh okay!
@twizmwazin Why don't we try to downgrade the command framework and see if that fixes the problem?

twizmwazin:
@dentmaged The issue isn't the version of the framework, the issue is the maven repo was down. All seems to be resolved.

Pablete1234:
@twizmwazin re-ammended the commit, still has issues

Pablete1234:
Tested with @twizmwazin and works

twizmwazin:
merged

PR #873, opened 2015-09-04
Fix /unmute

TiTo-418:


TiTo-418:
http://pastebin.com/PVHE9RYY

Pablete1234:
shouldnt this be localized?

CaptainElliott:
If you could localise this it would be great

PR #882, opened 2015-09-10
Implement a Teams are full message

Pablete1234:
If getTeamWithFewestPlayers is absent, a Teams are full error is sent


twizmwazin:
merged

Issue #553, opened 2015-04-06
Can exceed the max amount of players per team


TiTo-418:
E.G. Cannon Duel

RubeGamerV:
probably xml have [max-overfill="2"] or teams will be forced

AngelCasillas01:
@RubeGamerV On maps.oc.tc says 1, this is due to team force, we should reduce the power of the /team force

TiTo-418:
Isn't fixed

iEli2tyree011:
@TiTo-418 Can you give some more details for which case this isn't working in?

TiTo-418:
Sure, i ll make a video

SungMatt:
@TiTo-418 @iEli2tyree011 

Cannot reproduce.

Replication procedure assumed to be as followed:
1. Load a map with a low `max-overfill` tag
2. Join via `/join <team>` to fill a team
3. ???
4. Profit

Was not able to replicate issue.

[Video](https://www.youtube.com/v/0oWEXBUfI84)


*Tested in Cardinal version `00756e5`*

TiTo-418:
@SungMatt Try that in a non-blitz map.

TiTo-418:
Fixed with the team branch, i will reopen is happens again

TiTo-418:
<a href="http://imgur.com/iKAl0Dv"><img src="http://i.imgur.com/iKAl0Dv.png" title="source: imgur.com" /></a>

TiTo-418:
@Pablete1234 Could you explain the issue?

Pablete1234:
You can bypass the team "max" because that's the "non-vip" max, in cardinal everyone is treated as if they had vip, they can select their team and they can bypass the regular max, what they can't bypass is the "maxOverfill", and some maps don't even have one
What i think would solve this issue is that if maxOverfill is not set, maxOverfill is set to the same as max

dentmaged:
@Pablete1234 https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/module/modules/team/TeamModule.java#L52 and https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/command/JoinCommand.java#L47

That shouldn't happen though.

Pablete1234:
@dentmaged it should happen if it was public matches and you had a dux rank...

iEli2tyree011:
[...]

PR #885, opened 2015-09-12
Use max instead of maxOverfill when a player is joining a team

Pablete1234:
hopefully solves #553

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/team/TeamModule.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/team/TeamModule.java
@@ -49,7 +49,7 @@ public class TeamModule<P extends Player> extends ArrayList<Player> implements M
             player.sendMessage(new UnlocalizedChatMessage(ChatColor.RED + "{0}", new LocalizedChatMessage(ChatConstant.ERROR_MAY_NOT_JOIN, ChatColor.ITALIC + "" + ChatColor.AQUA + title + ChatColor.RESET + ChatColor.RED)).getMessage(player.getLocale()));
             return false;
         }
-        if (!force && size() >= maxOverfill) {
+        if (!force && size() >= max) {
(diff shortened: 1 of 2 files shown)


twizmwazin:
I'm pretty sure 1.2x is how Overcast handles it

Pablete1234:
@twizmwazin then we should change it so that players can't bypass the regular join limit, and if we are not going to use the maxOverfill, it may be kinda useless to have, maybe an op-only perm?

Pablete1234:
@twizmwazin  I changed it, but now it makes no sense to keep the maxOverfill anymore, this is a scrimm plugin, and I don't think we need permissions for maxOverfill, the regular max should be used

Also, about the 1.2x, I think that is how it works for public matches, but in tournaments (or any tm server) they dont do that because if you look at any HoC xml, they have "max=10" and no max overfill, but still when playing scrimms op's couldn't join to teams with 10 players in (or maybe op doesn't have the permission to join using "maxOverfill")

TiTo-418:
@Pablete1234 +1

Pablete1234:
@twizmwazin updated, max overfill is 25% more than regular max (https://docs.oc.tc/modules/teams)

Alfalik:
The additional 25 percent is for donator purposes. It's highly pointless have a ```max-overfill``` in Cardinal. I'd just set ```max-overfill``` & ```max``` to the same number.

Pablete1234:
the purpose of this PR is to use max instead of max-overfill, I don't know if twiz actually want me to get rid of everything related to max-overfill, or maybe do it a op-only feature?

TiTo-418:
I agree with @Alfalik

Pablete1234:
@TiTo-418 the thing is, why do you want 2 numbers that are the same? just get rid of one of them, i would remove max overfill overall (plugin would just ignore that in the xml)

twizmwazin:
I'm going to merge this with the justification that you can always change the team size. I will consider a permission for specific players to join to the max overfill.

PR #886, opened 2015-09-19
Teams now have a min players tag

Pablete1234:
It's a better system than the "if blitz, min 1 player per team" thing

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
@@ -8,7 +8,6 @@ import in.twizmwaz.cardinal.event.MatchStartEvent;
 import in.twizmwaz.cardinal.match.Match;
 import in.twizmwaz.cardinal.match.MatchState;
 import in.twizmwaz.cardinal.module.TaskedModule;
-import in.twizmwaz.cardinal.module.modules.blitz.Blitz;
 import in.twizmwaz.cardinal.module.modules.bossBar.BossBar;
 import in.twizmwaz.cardinal.module.modules.team.TeamModule;
 import in.twizmwaz.cardinal.settings.Settings;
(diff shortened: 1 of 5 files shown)


Pablete1234:
@dentmaged this tests if each team bypasses the min
the last ```if (count < minSize && !forced) {``` doesn't make sense, you are testing if number of teams that don't have enough players is smaller than the min players in total (for example 20),
doesn't make sense

dentmaged:
@Pablete1234 oh yeah, L55 should be count += team.size();

Alfalik:
The new ```min``` attribute also have a boss bar that show the quantity of player left to allow the match being started.

Pablete1234:
@twizmwazin where should i code a bossbar for the min tag? should i make a separate module? or should i put it in the start timer module?

twizmwazin:
@Pablete1234 yes, and as to where to put it, probably the start timer

Pablete1234:
@Alfalik I updated the PR and the "Waiting for 1 player to join" message shows up in the bossbar, but i feel like it's not quite optimized, (it's running every tick), what should i do about it? (you coded bossbars so you should know how to make it better)

src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
```diff
@@ -94,6 +92,12 @@ public void run() {
                 BossBar.delete();
             }
             time--;
+        } else if (GameHandler.getGameHandler().getMatch().getState().equals(MatchState.WAITING)) {
+            if (neededPlayers() > 0) {
+                BossBar.sendGlobalBossBar(waitingPlayerMessage(), 100F);
+            } else {
+                BossBar.delete();
```
Alfalik: You'll only delete the boss bar if ```neededPlayers() <= 0```. I'd also add an automatic start if the ```neededPlayers() <= 0``` with a default ```neededPlayers()``` of 20 players.

Alfalik:
@Pablete1234 If the boss bar need to be sent, a new wither will be created, after this, a check is run every tick. If a wither already exist, the health, location and name will be updated (we'll use the same wither) and if it does not exist, a new one will be created.

src/main/java/in/twizmwaz/cardinal/module/modules/startTimer/StartTimer.java
Pablete1234: I don't think we need automatic start, if both teams /ready, it will start
maybe i should make it so that if the /ready the countdown doesn't start if  all teams don't bypass the min tag

Pablete1234:
@Alfalik it would be nice if you could tell me what code you would put there

Alfalik:
You are always checking github, use that time to trial and error your code, instead of wasting it. Just to add, usually I don't use my free time to code (priorities)

Pablete1234:
@Alfalik this is not about trial and error, i just don't know what an efficient way of doing it would be

Alfalik:
The boss bar using a fake wither method is already implemented, just send the bar to the client. I don't see why you need to make it more "efficient"

Pablete1234:
because i don't know in what event i should send it, and as it is right now, i'm sending it every tick, and that is NOT efficient at all

Alfalik:
You are updating it (PacketPlayOutEntityMetadata packet), not sending new entities, (PacketPlayOutSpawnEntityLiving packet)

Pablete1234:
still, it's not efficient to update a wither each tick if it's not needed

dentmaged:
@Pablete1234 although you have a point, the server isn't under any heavy work before a game.

Pablete1234:
@twizmwazin set the needs testing tag, i think i fixed everything

Pablete1234:
Merged

PR #889, opened 2015-09-20
Fix Bossbar not dissapearing

Pablete1234:
reverts 7b9ec40 and makes a more localized fix, instead of forcing all withers to have atleast 1 health, because it messed up with bossbar.delete()


Prodigy4532:
Can i test this??

Alfalik:
Just check here https://github.com/Alfalik/CardinalPGM/commit/69dfab887340caaa113fea9cf2b924f35692531c

Alfalik:
I've did some further testing and this actually fix the bar not being deleted on cycle complete event, and also fix the monument modes bar still being visible after it runs. Hope this get merged soon.

twizmwazin:
merged

PR #897, opened 2015-09-27
Prevent pistons moving blocks into regions

### LINKED ISSUE #715: Pistons can push blocks outside the void regions
<a href="http://imgur.com/e6wc2Oi"></a>

For reference: https://github.com/twizmwazin/CardinalPGM/issues/559#issuecomment-104130726

Pablete1234:
fixes #715 
it also prevents blocks being retracted into void regions, and should pretty much fix all issues regarding pistons and regions
(tested in GDIII and blocks don't get pushed out of the map into the void)


Eudaldca:
Works as intended for me, I haven't noticed any bugs.

twizmwazin:
merged

PR #900, opened 2015-10-04
Crafting something with the enemy wool also destroys it

Pablete1234:
If you craft carpet or banners it will give you raindrops for destroying the enemy wool

known bugs:
 I can't get beds or paintings to work, as they do not have a set wool color


twizmwazin:
merged

PR #901, opened 2015-10-04
Fix cores leaking when they shouldn't

Pablete1234:
Bug:
If you break all the most bottom row of core blocks, the leak will be 1 block less, if you break all the blocks in the core, as soon as any of the lava flows, it will leak

Easy map to test in: regular SSB, break the bottom most pice ,and replace it with a block, then, break a block in the second layer, and it will leak
image


twizmwazin:
merged

PR #906, opened 2015-10-09
Allow joined players to teleport if the match is not playing

Pablete1234:
When waiting for a match to start, observers are allowed to tp, but joined players are not , that shouldn't work that way...


twizmwazin:
merged

PR #908, opened 2015-10-09
Change logic of messages at monument modes boss bar

### LINKED ISSUE #907: BossBar Issue | Time left for choose one class
https://gyazo.com/3ee5972725f90b79d892241d82b22df8

Alfalik:
Hope that anyone can confirm that this actually work and fix #907

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/monumentModes/MonumentModes.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/monumentModes/MonumentModes.java
@@ -104,9 +104,13 @@ public class MonumentModes implements TaskedModule {
                 }
             }
             if (timeBeforeMode > 0 && timeBeforeMode <= showBefore && name != null) {
-                int percent = (int) (timeBeforeMode * 100F) / showBefore;
-                if (timeBeforeMode <= 1) percent = 0;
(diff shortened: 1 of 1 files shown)


Pablete1234:
@Alfalik there is no need to make a localized message if it's just going to be a "{0}"
also, if the "translation" is the same for en and es, just put it in en.xml, it wil default to ther if it's not found on the es.xml

Alfalik:
@Pablete1234 thank you

Pablete1234:
@Alfalik np ;)

Pablete1234:
@Alfalik i think ```showBefore >= this.after``` is not what you should use, it just means if a mode is 5 mins in, and i say it should be showed 6 mins before, it shoudn't tell the seconds, that is not how it works, that case should either default to 60 secs or set showBefore to after, the bossbar should not display the "{0} in {1}" message, only if the mode is not going to occur in the match, so this.after > TimeLimit, or if it's more than 2^32 seconds (integer limit)

Alfalik:
What about the matches that does not have time limit? Like DATABasin

Pablete1234:
"or if it's a huge number, the thing is we don't know how huge it has to be"

Alfalik:
It isn't a big number, it's just basic logic and mathematics. It makes no sense to show a mode on the boss bar before 10 minutes and the mode is 5 minutes after the match starts. Does it?

@twizmwazin Any thoughts about the math-logic of my code?

Pablete1234:
@Alfalik the thing is, in databasing it's 1000h, show before 1000h, your rule would not work
if someone sets before 10 ,and mode is in 5, then asume it's an XML error and just set the before time to after (5 min)

Alfalik:
XML is not supposed to have errors or incoherent attributes, the show-before being the same as the after it's just a "trick"

Alfalik:
Does anybody knows another way to code this? Pretty sure this is how we should handle that "trick" used on DATABasin
```xml
<!-- 
Hack the boss bar to always show a string of text 
-->
<modes>
    <mode show-before="1000h" after="1000h" material="air" name="`6Change your class using `a/cl (name)"/>
</modes>
```
From: https://maps.oc.tc/DTM/Standard/DATAbasin/map.xml

Pablete1234:
@Alfalik this will have issues with matches that have no time limit, as the time limit is always 0 for matches with no time limit, it would never display the time

As i already told you before, the best way to handle this is only not to show time if the number of seconds is +3600, and it would be fixed

Pablete1234:
Fixed in 1.9

PR #915, opened 2015-10-25
Use regions also deny block break and block place events

Alfalik:
Here's an XML example:
```xml
<regions>
	<apply use="deny-all">
		<rectangle min="oo,oo" max="-oo,-oo"/>
	</apply>
</regions>


twizmwazin:
merged

Issue #911, opened 2015-10-11
Wildcard permissions don't work


Pablete1234:

TiTo-418:
Oh, now i know why refs couldn't manage the whitelist :c

angelitorb99:
Also you should be able to only put only " * " and recieve all permissions

PR #919, opened 2015-11-02
Add Admidistrator rank

elephantLocator:
Adds the rank of administrator to allow server owners not to have to op users. Has all cardinal permissions, as I believe wildcards work? They worked in testing at least.

This has been talked about on the issues page a bit, so I thought I might as well add it as it is currently existing on my server anyway. It's an alternative to opping people and having them run wild over all the server commands.

### DIFF
--- a/src/main/resources/ranks.xml
+++ b/src/main/resources/ranks.xml
@@ -35,4 +35,8 @@
         <permission>cardinal.punish.mute</permission>
         <permission>minecraft.command.pardon</permission>
     </rank>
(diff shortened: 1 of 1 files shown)


dentmaged:
Please fix the spelling mistake in the commit message with ```git commit -m "New message here" --ammend``` and ```git push -f```

Pablete1234:
@elephantLocator no, wildcard imports do not work #911 and it seems to work just because you have the moderator as parent, so the admin rank is mod perms + world edit

elephantLocator:
Can't figure that out. It's probably best to implement wildcards first anyway.

PR #925, opened 2015-11-25
Fix rightclick with book

Pablete1234:
ATM in cardinal in observers, if you are holding the book, you can interact with the world, for example, opening doors, tirggering buttons, and changig repeater delay

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/observers/ObserverModule.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/observers/ObserverModule.java
@@ -30,6 +30,8 @@ import org.bukkit.block.Dispenser;
 import org.bukkit.block.Dropper;
 import org.bukkit.block.Furnace;
 import org.bukkit.block.Hopper;
+import org.bukkit.craftbukkit.v1_8_R3.entity.CraftHumanEntity;
+import org.bukkit.craftbukkit.v1_8_R3.inventory.CraftItemStack;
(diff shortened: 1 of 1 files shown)


CaptainElliott:
Why not just prevent interaction all together instead of having a hard coded list of blacklsited blocks?

Pablete1234:
@CaptainElliott because then you can't open the book

twizmwazin:
Is it possible to cancel the event and then show the player the book? That just seems like a better solution to me.

Pablete1234:
@twizmwazin maybe it is (kinda like what you do with inventories) but i'm not sure, specially i'm not sure if i can show the contents properly, because they are not coded in atm

CaptainElliott:
Try using the `BlockPhysicsEvent` I'm pretty sure that is called when things like this happen

Pablete1234:
@CaptainElliott you can know that a block changed, but  you don't know who caused it, and therefore you can't cancel it if it's an obs doing it

Pablete1234:
@twizmwazin you can merge this already, it's fixed

twizmwazin:
Is Craftbukkit code the only way to fix this?

Pablete1234:
@twizmwazin it's the only way to "openInventory(Type.Book)", usually if you rightclick a button/trapdoor/whatever the "opening book interface" event is not triggered, and CraftBukkit i belive it's the only way to trigger it

twizmwazin:
merged

PR #927, opened 2015-11-26
Temporal fix prevent book right click

Pablete1234:
This just won't allow rightclick at all, the book can't be opened (but it doesn't matter atm, since the book contents are not coded anyways)


Pablete1234:
The real fix is ready to merge, no need for a temporal fix anymore

PR #928, opened 2015-11-28
Prevent projectiles hitting yourself, fixes #855

### LINKED ISSUE #855: You can damage yourself with your arrows
Latest [versions](http://prntscr.com/8970xq)

Pablete1234:
It still plays the *ding* sound if you hit yourself, but it's 99.9999% less annoying than before

Note: removed the metadata as i felt it's not necesary, setting friendlyfire on the teams fixes hitting other players, it only doesn't work on yourself
Tested with 2 accounts, you can hit players in other teams, you can't hit yourself, and you can't hit your own team

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/friendlyFire/FriendlyFire.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/friendlyFire/FriendlyFire.java
@@ -10,12 +10,11 @@ import in.twizmwaz.cardinal.util.Teams;
 import org.bukkit.entity.Arrow;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
+import org.bukkit.entity.Projectile;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.HandlerList;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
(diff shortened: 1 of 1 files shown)


Prodigy4532:
You can blow your own team.

Pablete1234:
@Prodigy4532 i belive that's controled separetly, and the map decides if it's enabled or disabled

Pablete1234:
@Prodigy4532 https://github.com/twizmwazin/CardinalPGM/blob/master/src/main/java/in/twizmwaz/cardinal/module/modules/disableDamage/DisableDamage.java is what controls TNT, and it's not quite as good as it should be, it should handle everything here:  https://docs.oc.tc/modules/damage

Pablete1234:
@twizmwazin opinions? it removes all the other code, but it works, and does exactly what it needs to do. other players in your same team don't get hit because team friendly fire is set to off.

twizmwazin:
Aren't you supposed to be able to hit yourself in PGM?

Pablete1234:
@twizmwazin not with arrows, they changed sportbukkit so that frendly fire setting on the team does not affect yourself, and in PGM for some days you could shoot yourself, but that got fixed, and this also needs a fix.

ps: this does not affect tnt, your own tnt does kill yourself

Pablete1234:
merged

PR #929, opened 2015-11-28
Permissions help

unipablojorge:
I want to know what the permission is to be able to join a team


Alfalik:
The ```/join``` and ```/leave``` commands does not have permissions (all players can execute the command), because Cardinal purpose are scrimmages. Also this is not the correct way to ask doubts, you should simply create an issue https://github.com/twizmwazin/CardinalPGM/issues/new or check next time the wiki https://github.com/twizmwazin/CardinalPGM/wiki.

Pablete1234:
why is this a PR in the first place?

CaptainElliott:
refer to Alfalik's (comment)[https://github.com/twizmwazin/CardinalPGM/pull/929#issuecomment-160295943]

PR #934, opened 2015-12-07
Long distance render tnt

Pablete1234:
Needs testing, and i'm not sure about how this affects performance with a big amount of players & tnt

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
@@ -34,6 +34,7 @@ import in.twizmwaz.cardinal.module.modules.itemRemove.ItemRemoveBuilder;
 import in.twizmwaz.cardinal.module.modules.killReward.KillRewardBuilder;
 import in.twizmwaz.cardinal.module.modules.killStreakCount.KillStreakBuilder;
 import in.twizmwaz.cardinal.module.modules.kit.KitBuilder;
+import in.twizmwaz.cardinal.module.modules.longTntRender.LongTntRenderBuilder;
 import in.twizmwaz.cardinal.module.modules.mapNotification.MapNotificationBuilder;
 import in.twizmwaz.cardinal.module.modules.match.MatchModuleBuilder;
 import in.twizmwaz.cardinal.module.modules.matchTimer.MatchTimerBuilder;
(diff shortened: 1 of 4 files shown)


TiTo-418:
https://youtu.be/ibI4F2GTNKQ

TiTo-418:
https://youtu.be/ibI4F2GTNKQ

Pablete1234:
Real scrimm test: http://www.youtube.com/watch?v=fp6IA3ZzFeE

TiTo-418:
https://youtu.be/ibI4F2GTNKQ

Pablete1234:
New testing: https://youtu.be/dFLmknc8fDw
IMO it's pretty much done

twizmwazin:
As soon as 1.9 is merged I want to get this on master

Pablete1234:
@twizmwazin in 1.9 some entity rendering changed, the render distance for tnt is now half the client chunk render distance, and armor stands now render as far as your render distance goes, so i would rather use them than fake players

twizmwazin:
That's fine

Pablete1234:
@twizmwazin the issue is, how far away tnt should render dependes on the client settings, so having long render distance tnt is a bit harder, as we need to get the data from the client settings packet to know the amount of chunks the player is rendering

twizmwazin:
I believe there is a way of getting some of players' settings. The server uses it to know how many chunks to send to the client.

Pablete1234:
@twizmwazin yes, the setting exists, and it's sent by the client in the client setting packets, but that packet doesn't have an event, and doesn't get saved anywhere (that i'm aware of) in NMS (i tried to find it in 1.8 for the wither)

Pablete1234:
Fixed and updated for 1.10, missing the setting for chunks, default to 64 blocks (as if all players had 8 chunks, if playing with more than 8 chunks, you will see the tnt twice (the real tnt and the fake tnt) in specific distances).
Videos of it in action (with 8 chunks render distance) :
https://youtu.be/xzaCbt5FtFI
https://youtu.be/4LVwyeOQxrs

Pablete1234:
It's now working, and it reads player's chunks, but 2 things to keep in mind:
 - You need SportBukkit with OvercastNetwork/SportBukkit#210 merged to run this.
 - There is currently a bug in the minecraft client, where it won't tell the server you changed your render distance if you use esc key to exit the settings menu (make sure you press "done") [#MC-108315](https://bugs.mojang.com/browse/MC-108315).

PR #935, opened 2015-12-08
Add material as filter case

Pablete1234:
in 1.4.0 <block> was replaced by <material>, in cardinal we shouldn't remove block, as it doesn't support multiple prot., but we should support material


twizmwazin:
Merged

PR #937, opened 2015-12-14
Fix inventory view not opening if the display name is too big

Pablete1234:
The max title length for an inventory title is 32 characters, if the player's name + ranks is bigger than 32 characters, it will throw an exception and not open the inventory


twizmwazin:
Merged

PR #943, opened 2015-12-18
Implement multitrade

Pablete1234:
Title says all the needed stuff, and the video shows how it works:
https://www.youtube.com/watch?v=U97-kmSA3U4

(it still has some bugs throwing npe's when cycling etc, but i will try to fix that)

ps: enjoy the 2k


twizmwazin:
Looks good so far

twizmwazin:
merged, formatting will need to be fixed in master

PR #945, opened 2015-12-19
Allow id & name in filters and regions, add everywhere and nowhere regions, and default apply region to everywhere, fixes #723

### LINKED ISSUE #723: Sumo doesn't work
Temporarily fix: http://pastebin.com/zTA5xJr4 by @Pablete1234 

What is sumo?: https://www.youtube.com/watch?v=alJB0ycrWb0
Proof: http://youtu.be/2J56zEqTACg
XML: https://maps.oc.tc/Other/Tournament/House%20of%20Cores/Sumo/map.xml

Pablete1234:
this PR atm fixes completly all the bugs that made sumo not work (when talking about sumo, i'm also talking about general 1.4 proto maps), but maybe it broke something in 1.3 (i don't think so), but i would like to test this a bit.


twizmwazin:
merged

PR #946, opened 2015-12-20
Book gets message from CardinalNotifications

Pablete1234:
This will download the book from the CardinalNotifications repo and give it to the player
This will allow the book to be updated directly from github without the need of a new cardinal build to update it

I renamed the "UpdateNotification" module to "CardinalNotifications"


Pablete1234:
merged

PR #949, opened 2015-12-24
List winning teams in order

Pablete1234:
This PR replaces the old get who won system to get a list from it, so that you can sort the scoreboard properly in maps with more than 2 teams, also this PR uses closest completion instead of total uncompleted points
Also this system takes into acount everything, for example, if teams tie in closest proximity, it will go check for the second proximity.


Pablete1234:
merged

PR #950, opened 2015-12-27
Implement fancy tab

Pablete1234:
The feature is fully working except that it's missing skins, everything else works, even tho i would like to remove the code from the header module, and create it's own module, also some events that update it may be missing

PS: any help implementing skins will be nice

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/Cardinal.java
+++ b/src/main/java/in/twizmwaz/cardinal/Cardinal.java
@@ -41,6 +41,7 @@ import in.twizmwaz.cardinal.rank.Rank;
 import in.twizmwaz.cardinal.rotation.exception.RotationLoadException;
 import in.twizmwaz.cardinal.settings.Setting;
 import in.twizmwaz.cardinal.settings.SettingValue;
+import in.twizmwaz.cardinal.tabList.TabList;
 import in.twizmwaz.cardinal.util.ChatUtil;
 import in.twizmwaz.cardinal.util.DomUtil;
 import in.twizmwaz.cardinal.util.Numbers;
@@ -50,6 +51,7 @@ import org.bukkit.command.Command;
 import org.bukkit.command.CommandSender;
 import org.bukkit.command.ConsoleCommandSender;
 import org.bukkit.configuration.file.FileConfiguration;
+import org.bukkit.event.Listener;
 import org.bukkit.plugin.java.JavaPlugin;
 import org.jdom2.Document;
 import org.jdom2.Element;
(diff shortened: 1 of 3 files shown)


TiTo-418:
/team alias doesn't change the name in the tab

There are others bugs like this one http://imgur.com/09Eygy1

Pablete1234:
@TiTo-418 i already fixed the team join updating

Eudaldca:
@Pablete1234 Minor, but some tab UUID's match real player's UUID.

Pablete1234:
how should i go arround it?

Pablete1234:
This should be ready to use in all servers, but may have issues if the server has +70 players (even tho it shouldn't)

Pablete1234:
Only known bug atm is that hats don't show up (skins show up, but not the overlay)

Eudaldca:
Great job @Pablete1234! :clap:

dentmaged:
@Pablete1234 look at this: http://dentmaged.github.io/docs/sportbukkit/org/bukkit/Skin.html

Pablete1234:
@xFracTV already known bug, and i have a patch for it, i'll push it later

Pablete1234:
@xFracTV fixed

twizmwazin:
Made a few notes on formatting, this is looking really great though.

xFracTV:
when you use " ". Team name does not appear

Pablete1234:
I knew you could "/team alias blue \u00A74Red" and it would use unicode values, didn't think quotes would give problems
@twizmwazin what is the best way i could escape that?
```
L329: IChatBaseComponent.ChatSerializer.a("{text:\"" + displayName + "\"}")
```
means:
```
{text:" + whatever + "}
```

twizmwazin:
I think you figured it out. Alternately, we could prohibit certain chars from being used in team names

Pablete1234:
I don't think we need to prohibit any characters, i only need to escape the characters that go into the text, my question is, what is the best method to escape a string?

Pablete1234:
I don't think we need to prohibit any characters, i only need to escape the characters that go into the text, my question is, what is the best method to escape a string?

twizmwazin:
Backslashes are the only way to escape strings

Pablete1234:
I know that, but what method/library/util/whatever does it with all chararters that need it?

CaptainElliott:
@Pablete1234 try out apache commons https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringEscapeUtils.html#escapeJava(java.lang.String)

Pablete1234:
moved on to fancy tab V2 #956

PR #952, opened 2015-12-30
Safer module builder loading

CaptainElliott:
Removes the unsafe and bad raw type issue. Should work fine haven't done extensive testing

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/ModuleFactory.java
@@ -80,100 +80,108 @@ import in.twizmwaz.cardinal.module.modules.worldFreeze.WorldFreezeBuilder;
 import org.bukkit.Bukkit;
 
 import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.ArrayList;
 import java.util.logging.Level;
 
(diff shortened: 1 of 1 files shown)


Pablete1234:
@CaptainElliott i thought you were dead

CaptainElliott:
I came back its ok... 

Regarding PR just read through it as I did it in a rush as I needed to leave my house, I accidentally instantiated the set in the variable declaration instead of the contructor. 

I also think I might rename the method to something lie addBuilders.

In addition now that this removes the raw type array and replaces it with a parameterlized set I guess the list could me removed? 

That all needs to be done before this can even be considered to merge.

CaptainElliott:
Probably should have waited to get home to do this PR, anyway also just wanted to add some more description to the point of the PR. 

It removes the unsafe raw type sclera room for classes which could throw a ClassCastException at runtime if a mistake was made in the list of builders resulting it to contain a non builder class. Instead with the parameterlized type it will give a warning at compile time - which is obviously much better.

Sorry if there's any mistakes in these comments - on phone

twizmwazin:
I'm not a fan of wildcard imports.

CaptainElliott:
@twizmwazin will fix when I'm back home

CaptainElliott:
@twizmwazin done, can squash if you think this is all ok

CaptainElliott:
@twizmwazin all squashed for ya

CaptainElliott:
@twizmwazin merge?

twizmwazin:
done

PR #954, opened 2016-01-12
Changing the method of disabling perms

angelitorb99:
To disable a permission that is given to default players you have to use that method.


CaptainElliott:
What's wrong with unsetting the permission? 
I mean you're disabling the permission from being used for that rank.. so you should be removing it from the 1PermisisonAttachment` not setting it to false

twizmwazin:
I'm going to close this as I don't think this does anything.

PR #955, opened 2016-01-15
Allow players to join obs even if the team changed name

Pablete1234:
/leave should always join obs.
I made it so that if you /join, and no team is found, and you typed part of the word observers, it will join you to obs, so, all : "/team o", "/team obs", /team "obser" should join you to obs even if the name was changed, unless there is a team that starts with that ("/team o" would join you to orange if there is a team named orange)


Zenzoh:
CardinalPGM needs a AntiCheat, https://oc.tc/_RickAstley is a good option, but he's can detect all the actual Ghost Clients, AutoClicks, etc...

Please, Pablete1234 or anyone create a pull request with this code:

"ONDA VITAL RICKERINO PAN Y VINO"

twizmwazin:
merged

PR #958, opened 2016-01-31
Fix null header names, display flairs for offline players, add map author flair, fixes #736

### LINKED ISSUE #736: inconsistency with tab header names and cycled maps descriptions
differences in maps:
http://gyazo.com/16831d004a9511bc733e2a72cd454ffc

maps with correct color get player "null", and the maps that for some reason load with old color, have the name right (both maps use UUID's on their xml's)
the difference is that fractal (name = null) uses:
```<author uuid="dbe909c6-2b15-4528-b12a-c77ed4031496" contribution="Design and building"/>```
and Golden Drought III uses:
```<author uuid="459fd919-0d58-401a-9601-6cab1eb721ab"/>```

also this http://gyazo.com/21973f36ca69a00d391fb711720acaeb should have the new names

ps:
also adding to this, if the map creator is in the game, they will always be displayed as the name they had on the last match (joined red or blue or obs from the last match), because that only gets set on map load, not on player joining teams event

Pablete1234:
Also updates the /map format to:
image
(Only talking about authors and contributors)
gamemode(needs 1.4 protocol), edition (N/A in a scrimm server) and XML(xml's are local, can't be viewed in a webpage) are not implemented in cardinal.

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/chat/ChatConstant.java
+++ b/src/main/java/in/twizmwaz/cardinal/chat/ChatConstant.java
@@ -202,6 +202,7 @@ public enum ChatConstant {
     GENERIC_MUTED_BY("generic.mutedBy"),
     GENERIC_UNMUTED("generic.unmuted"),
     GENERIC_UNMUTED_BY("generic.unmutedBy"),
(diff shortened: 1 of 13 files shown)


Pablete1234:
merged

PR #963, opened 2016-02-14
Rewrite firework system, fixes #909

### LINKED ISSUE #909: Fireworks
In "Overcast Network" Fireworks Work different, when you complete a monument 4 fireworks are launched around the monument. In "CardinalPGM" is released only 1 and the player who complete the monument.

Pablete1234:
it will spawn fireworks arround the objective instead of one at the position of the player


CaptainElliott:
I would abstract away the `getRadius` method

Pablete1234:
@CaptainElliott i tried to, and i had something like:
Before the firework:
```
Bukkit.getConsoleSender().sendMessage("Vector:" + region.getSize());
```
On the region:
```
@Override
    public Vector getSize() {
        Vector vec = max.clone().subtract(min);
        Bukkit.getConsoleSender().sendMessage("Vector will be:" + vec);
        return vec;
    }
```
And the output was:
```
Vector will be: 4.0,4.0,4.0
Vector: Infinity, Infinity, Infinity
```

I didn't find a way arround (and i asked twiz and he didn't know why either), so i just made it like that

Pablete1234:
Implemented in #993

PR #964, opened 2016-02-16
Fix wither blinking for the missing players bossbar

Pablete1234:
I also removed some innecesary code, but i want to test this a bit before i merge, just in case


twizmwazin:
Is this ready?

Pablete1234:
@twizmwazin i've been testing this in a couple servers for some days and didn't seem to keep wither alive or anything like that, so i think it's ready

twizmwazin:
merged

PR #965, opened 2016-02-18
Fix empty result displayed on snowflakes command.

iPGz:
When the `snowflakes` command is run and the player specified has no raindrops (stored as `""`) the message will return with a blank space (pulling from the empty db entry). This fix checks the snowflake amount previous to creating the ChatMessage and uses the value returned within the building of the message (translating `""` to `0`).

- Display a String of `0` if the return value is null/`""` (player has no
snowflakes).

- Reduced to only check the db only once.


Pablete1234:
+1

CaptainElliott:
Good for merge :+1:

twizmwazin:
merged

PR #972, opened 2016-03-05
Add Dispenser TNT tracking.

iPGz:
- Dispensed TNT is tracked using a HashMap of placed dispensers.

The source of the TNT is set to be the same owner as the dispenser.


CaptainElliott:
:+1:

twizmwazin:
merged

PR #982, opened 2016-03-17
Kits rewrite

Pablete1234:
This is done ontop of the 1.9 rebase

This replaces the old 1 kit that applies everything system, for an interface kit, that has the apply function, and lets diferent kit types handle it as they should, the "main" kit, the one that gets referened by the XML, is a KitCollection (it applies all it's kit types, and parent KitCollections)

Still in Development, but basics (items and armor) work


Pablete1234:
#983

Issue #980, opened 2016-03-16
Potion module not implemented


Lmmb74:
Hi, I found that the modules potion, void and the new kit module are not implemented in Cardinal, could somebody add them?

Pablete1234:
@Lmmb74 `<above y="90"/>` half space regions are not implemented

Lmmb74:
Thanks

Lmmb74:
Pablete, necesito que me ayudes en algo, si tenes skype contactame por favor, mi usuario es Lmmb74 en skype

twizmwazin:
@Lmmb74 This issue tracker is english-only. If you are going to say something in your native language, provide an english translation, or at least your best attempt.

Lmmb74:
Ok, I was telling Pablete to contact me on skype, since I need his help. And sorry.

Pablete1234:
Fixed with Kits Rewrite

Issue #745, opened 2015-07-01
double jumping not implemented


Pablete1234:
Tips to implement:
oc.tc detects double jumping using the "canFly" nbt on all players with a kit than can double-jump, and if flying=1, they set flying=0 and apply the velocity

twizmwazin:
We need to find the specific vector that is applied, it might be on docs. Also, unless you have only a single finger, please make an attempt at capitalization, it makes it easier to read and looks more professional.

Pablete1234:
isn't the vector defined by the map in xml?

CaptainElliott:
@Pablete1234 No

TiTo-418:
https://docs.oc.tc/modules/kits

dentmaged:
@twizmwazin @TiTo-418 The `power` is `3`, `recharge-time` is `2.5s` and `recharge-before-landing` is `false`.

twizmwazin:
@dentmaged Thanks for reading Tito's screenshot for me. It still doesn't have the vector that I need.

TiTo-418:
@twizmwazin What do you mean whit "vector"?

CaptainElliott:
@TiTo-418 velocity

dentmaged:
@twizmwazin oh lol, I didn't see that screenshot.

Pablete1234:
@CaptainElliott you are a map developer, can't you create a dev map with double jumping power 1, and always looking up test /vel 0 ? 0 , and the double jump, and keep testing until you find out what the motion applied to the player is when double jumping with power 1

TiTo-418:
Maybe this can help https://repo.oc.tc/content/repositories/snapshots/tc/oc/double-jump/

dentmaged:
@twizmwazin @TiTo-418 https://github.com/rmsy/DoubleJump

TiTo-418:
'''
Vector normal = VectorUtil.calculateLookVector(player.getLocation());

            normal.setY(0.75 + Math.abs(normal.getY()) * 0.5);
            event.getPlayer().setVelocity(normal);

            player.getWorld().playSound(player.getLocation(), Sound.ZOMBIE_INFECT, 0.5f, 1.8f);
'''

TiTo-418:
' ' '
test
' ' '

Pablete1234:
Fixed with Kits Rewrite

PR #983, opened 2016-03-17
Kits rewrite

Pablete1234:
This is done ontop of the 1.9 rebase

This replaces the old 1 kit that applies everything system, for an interface kit, that has the apply function, and lets diferent kit types handle it as they should, the "main" kit, the one that gets referened by the XML, is a KitNode (it applies all it's kit types, and parent KitNodes)

All kits should now (Except shield kit, but i didn't find a single map that uses it)

fixes:
 - `<effect>` tag on kits and items #980
 - Doble jump not implemented #745


src/main/java/in/twizmwaz/cardinal/module/modules/kit/Kit.java
```diff
@@ -1,163 +1,11 @@
...
-                player.getInventory().addItem(bookStack);
-            }
-        }
-    }
-
-
-    public String getName() {
-        return name;
-    }
+    public void apply(Player player, Boolean force);
```
twizmwazin: You don't need to mark this as public

src/main/java/in/twizmwaz/cardinal/module/modules/kit/KitCollection.java
```diff
@@ -0,0 +1,101 @@
...
+import org.bukkit.event.HandlerList;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class KitCollection implements KitRemovable {
+
+    private String name;
+    private FilterModule filter;
+    private Boolean force;
```
twizmwazin: You probably mean to use the primitive type boolean here, not Boolean

src/main/java/in/twizmwaz/cardinal/module/modules/kit/KitCollection.java
```diff
@@ -0,0 +1,101 @@
...
+import org.bukkit.entity.EnderPearl;
+import org.bukkit.entity.Entity;
+import org.bukkit.entity.Player;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.HandlerList;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class KitCollection implements KitRemovable {
```
twizmwazin: I'm not sure that KitCollection is the best possible name here. It implies that it extends Java's Collection.

src/main/java/in/twizmwaz/cardinal/module/modules/kit/kitTypes/ClearKit.java
```diff
@@ -0,0 +1,33 @@
+package in.twizmwaz.cardinal.module.modules.kit.kitTypes;
+
+import in.twizmwaz.cardinal.module.modules.kit.Kit;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+public class ClearKit implements Kit {
+
+    Boolean clear;
```
twizmwazin: primitive type?

src/main/java/in/twizmwaz/cardinal/module/modules/kit/kitTypes/HealthKit.java
```diff
@@ -0,0 +1,31 @@
+package in.twizmwaz.cardinal.module.modules.kit.kitTypes;
+
+import in.twizmwaz.cardinal.module.modules.kit.Kit;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+
+public class HealthKit implements Kit {
+
+    int health;
```
twizmwazin: private

src/main/java/in/twizmwaz/cardinal/module/modules/kit/kitTypes/ItemKit.java
```diff
@@ -0,0 +1,72 @@
...
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.PlayerInventory;
+
+import java.util.List;
+
+public class ItemKit implements Kit {
+
+    List<KitItem> items;
```
twizmwazin: private

src/main/java/in/twizmwaz/cardinal/module/modules/kit/kitTypes/PotionKit.java
```diff
@@ -0,0 +1,37 @@
...
+import in.twizmwaz.cardinal.module.modules.kit.KitRemovable;
+import org.bukkit.entity.Player;
+import org.bukkit.event.HandlerList;
+import org.bukkit.potion.PotionEffect;
+
+import java.util.List;
+
+public class PotionKit implements KitRemovable {
+
+    List<PotionEffect> effects;
```
twizmwazin: private

src/main/java/in/twizmwaz/cardinal/module/modules/kit/KitCollection.java
Pablete1234: I don't think it makes sense to extend/implement collection, it would be better to find another name, but any name i can think of, that represents a "group of kits" seems like it's an extension of "group" (like KitArray, or KitList) :/

twizmwazin: A collection is just a group of zero or more objects.

Pablete1234: But the KitCollection has a List<Kit> and a List<KitCollection> (parents), what one should i apply the stuff on? it's kinof wierd, it should have it's own methods, not inherit anything from collection

twizmwazin: If you implement the collection interface, you define all the methods.

Pablete1234: I know i define all the methods, but i have to set all the ones that the Collection interface has, and i don't really feel like it's needed at all
Methods like contains, add/remove, toArray, stuff like that will most likely never get used in the KitCollection class, and if you want to do it, you need to specify if you want to do it on the subkits or on the parent KitCollection list

twizmwazin: I still feel that if you're gonna call it a collection it needs to be a collection.

Pablete1234: I know that, and i want to change the name, the question is, what should i change it to?

twizmwazin: KitComplete?

Also, perhaps consider moving the boolean values to their own kit objects, it makes the most sense to fully modularize it. It may also be worth it to store actual references to the parent kits rather than just a string.

Pablete1234: I do keep a refernce to the parent kits, the issue is you may try to get a parent kit that didn't get created yet, it stores a string, and on cycle complete, it uses the string to find the parent / parents, and stores them in the parentKits list

Pablete1234: If you look at https://docs.oc.tc/modules/kits, the 3 values that i kept as booleans in the parent kits are there because they will always be set to what you set them to, or to the default, the idea of making kit subclases, is so that it adds the posibility to make one do stuff, but don't have the other one do anything

A perfect example was the walk speed kit, in the old system, walk speed was set to 0.2F by default, and changed to something else if specified, well, the kit didn't work on boomBox / ender blast because it was first applying the modified kit, and afterwards (parent kit) applied default speed again

TiTo-418: Maybe KitNode?

src/main/java/in/twizmwaz/cardinal/module/modules/kit/kitTypes/AttributeModifierKit.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public void unload() {
+        HandlerList.unregisterAll(this);
+    }
+
+    @Override
+    public void apply(Player player, Boolean force) {
+        // TODO
+        // How do you apply an attribute to a player?
```
Pablete1234: Anyone has any idea?

Pablete1234:
Merged into rebase1.9 branch

PR #988, opened 2016-03-24
Titlerespawn

Pablete1234:
I belive i fixed most if not all the bugs that this had, and rebased it to 1.9
I would specially like testing in the visibility part (if dead players can be seen, or if dead players can see obs, stuff like that), because i didn't test that too much yet


src/main/java/in/twizmwaz/cardinal/module/modules/armorKeep/ArmorKeep.java
```diff
@@ -71,6 +71,7 @@ public void onPlayerDeath(PlayerDeathEvent event) {
 
     @EventHandler(priority = EventPriority.HIGHEST)
     public void onPgmSpawn(CardinalSpawnEvent event) {
+        if (event.isCancelled()) return;
```
twizmwazin: I believe all of these can be replaced with ignoreCancelled = true in the annotation

src/main/resources/lang/en.xml
```diff
@@ -366,4 +362,19 @@
         <snowflake>snowflake</snowflake>
         <snowflakes>snowflakes</snowflakes>
     </snowflakes>
+    <death>
```
twizmwazin: this looks like a mess

Pablete1234: It definetly looks like a mess, but it's the order / paths that ocn uses, so it's the best way to implement that imo

twizmwazin: We don't closely follow ocn's pattern elsewhere, I don't see a binding reason to here.

Pablete1234: The reason is ocn's xml will specify the death message translations with that paths, i can remove the click messages and that, but the waiting messages should stay with that path

twizmwazin: Don't worry about ocn's paths here. Make a conversion method to translate if you need to. We have a whole new system coming for cardinal2 when that gets out the door.

Pablete1234:
Changing to a PR to master

PR #989, opened 2016-03-25
Titlerespawn

Pablete1234:
I belive i fixed most if not all the bugs that this had, and rebased it to 1.9
I would specially like testing in the visibility part (if dead players can be seen, or if dead players can see obs, stuff like that), because i didn't test that too much yet


Pablete1234:
IMO this is pretty much ready to be released, only waiting for bug reports, (specially visibility issues)

twizmwazin:
merged

PR #992, opened 2016-03-29
Add item mods module

Pablete1234:
This is needed to apply all changes that ocn made to combat in 1.9 (they add item attributes to all armors to balance them)


src/main/java/in/twizmwaz/cardinal/util/Parser.java
```diff
@@ -6,6 +6,7 @@
 import org.apache.commons.lang.math.NumberUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
+import org.bukkit.Bukkit;
```
twizmwazin: unused?

Pablete1234: Most likely, i added it for some debug messages

PR #993, opened 2016-04-04
Implement koth, capture points

Pablete1234:
Tested with 5cp (Ascendance 2 & smoke) and works as expected, may need some changes for different protocols

getMin and getMax functions in regions are intended for the fireworks PR, and to try implement the saving of blocks (¿bukkit BlockImages / Chunk snapshots?) before the regions get changed (Atm i'm setting the wool/clay/stained glass to either 0 (white) when there's no team, or the team's color, it should be either the team's color, or the block that comes placed in the map)

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/hill/HillObjectiveBuilder.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/hill/HillObjectiveBuilder.java
@@ -1,21 +1,24 @@
 package in.twizmwaz.cardinal.module.modules.hill;
 
+import com.google.common.collect.Sets;
 import in.twizmwaz.cardinal.match.Match;
-import in.twizmwaz.cardinal.module.BuilderData;
 import in.twizmwaz.cardinal.module.ModuleBuilder;
 import in.twizmwaz.cardinal.module.ModuleCollection;
-import in.twizmwaz.cardinal.module.ModuleLoadTime;
 import in.twizmwaz.cardinal.module.modules.filter.FilterModule;
 import in.twizmwaz.cardinal.module.modules.filter.FilterModuleBuilder;
+import in.twizmwaz.cardinal.module.modules.filter.parsers.BlockFilterParser;
+import in.twizmwaz.cardinal.module.modules.filter.type.BlockFilter;
+import in.twizmwaz.cardinal.module.modules.filter.type.logic.AnyFilter;
 import in.twizmwaz.cardinal.module.modules.regions.RegionModule;
 import in.twizmwaz.cardinal.module.modules.regions.RegionModuleBuilder;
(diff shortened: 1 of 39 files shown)


src/main/java/in/twizmwaz/cardinal/module/modules/hill/HillObjectiveBuilder.java
```diff
@@ -23,103 +29,79 @@
...
                 }
             }
         }
         return results;
     }
 
-    private HillObjective parseHill(Element element) {
-        String id = element.getAttributeValue("id");
-        String name = element.getAttributeValue("name");
+    private HillObjective parseHill(Boolean king, Element... elements) {
```
twizmwazin: boolean, not Boolean

twizmwazin:
formatting everywhere: methods should have a blank line between them

Pablete1234:
@twizmwazin I didn't event realize getMin and getMax were not separated by a line, and i copy-pasted the methods all thru the regions, it's fixed

src/main/java/in/twizmwaz/cardinal/module/modules/hill/RegionSave.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class RegionSave {
+
+    private final BlockVector min;
+    private final BlockVector size;
+    private final List<Pair<Material,Byte>> blocks;
+
+    RegionSave(RegionModule region) {
```
twizmwazin: package

src/main/java/in/twizmwaz/cardinal/module/modules/score/ScoreModule.java
```diff
@@ -71,11 +71,21 @@ public void unload() {
     }
 
     public int getScore() {
```
twizmwazin: Might we want to return a double here and then cast to an int when we need to?

Pablete1234: Well, i just let it like that, because it's only really used in the scoreboard, or chat, where it always displays as an int, also, i don't think all arround the code we should use "setScore(getScore() + 1)", and call the score event, it's way better to "addScore(something)", and if it's different than the old socre, it calls the socre event

src/main/java/in/twizmwaz/cardinal/module/modules/hill/RegionSave.java
Pablete1234: twiz, this might be used for renewables in the future, that's why i didn't really add a package

twizmwazin: I'm saying that it isn't public/private/protected. Was this intentional?

Pablete1234:
merged

PR #994, opened 2016-04-04
Add attack speed and damage indicators

Pablete1234:
Changes:
 - Attack speed to x5
 - Arrows don't get affected by player velocity, and they no longer have a +- random angle
 - Added floating numbers indicator (On by default switch with `/toggle damage`)
 - Added attack speed indicator on the action bar (Off by default switch with `/toggle cps`)
 - Enlarges player's hitboxes to be 0.75 wide (+25% than default)

I made it so when server starts up, it always copies the settings from the config.yml inside the jar, this will prevent old bugs that made it so new toggles never got added if you didn't clear the config


Pablete1234:
@twizmwazin this is really ready to release, check if you can see anything wrong, and i'll merge it

twizmwazin:
Wouldn't always copying the config override user settings every time the server starts?

Pablete1234:
@twizmwazin yes it does override the settings, but i've never knew anyone that edited the settings in any server (if really necesary, they can startup the server, then change the settings, and run /cardinal -r),  in some servers, you still cant /toggle pms, bevause of an old config, and i think this is the best solution

twizmwazin:
If the settings can't be changed, they shouldn't be in the settings. This cannot be merged until that is fixed.

Pablete1234:
@twizmwazin moved settings to it's own settings.yml, (i let the lines of code that clear them from the main config, as well as a line that clears ranks from main config, they were moved to the database)

src/main/java/in/twizmwaz/cardinal/module/modules/damageIndicator/PlayerBoundingBox.java
```diff
@@ -0,0 +1,183 @@
...
+import org.bukkit.event.player.PlayerMoveEvent;
+import org.bukkit.event.player.PlayerTeleportEvent;
+import org.bukkit.event.player.PlayerUseUnknownEntityEvent;
+import org.bukkit.util.Vector;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+public class PlayerBoundingBox implements Listener {
```
twizmwazin: Is there a reason we decided to make everything package-level in this class?

Pablete1234: What does that mean?

twizmwazin: not public, private, or protected

ElliottWhyman: Has no acess modifier meaning that it can be used inside it's class aswell as any other classes in the same package **only**

Pablete1234: @ElliottWhyman yeah, i know what it does, i just didn't understead twiz's question

Pablete1234:
@twizmwazin Fixed package-level access, ready to merge?

src/main/java/in/twizmwaz/cardinal/module/modules/damageIndicator/DamageIndicator.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.UUID;
+
+public class DamageIndicator implements Module {
+
+    public static double width = 0.75;
+    public static double zombieHitBox = 0.6;
+
+    public static double offset = (width - zombieHitBox) / 2;
+
+    HashMap<UUID, PlayerBoundingBox> boundingBoxes = new HashMap<>();
```
twizmwazin: package

src/main/java/in/twizmwaz/cardinal/module/modules/damageIndicator/DamageIndicator.java
```diff
@@ -0,0 +1,152 @@
...
+import org.bukkit.util.Vector;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.UUID;
+
+public class DamageIndicator implements Module {
+
+    public static double width = 0.75;
```
twizmwazin: these fields should be made final constants. Constant names are usually written in caps separated by underscores

twizmwazin:
Not sure if you forgot to add it to git, but the fields are still package

Pablete1234:
@twizmwazin I fixed more stuff, changed package-level, and fixed it to work with title respawn (you could hit a player while dead, because of the player hitbox thig)

src/main/java/in/twizmwaz/cardinal/module/modules/damageIndicator/PlayerBoundingBox.java
```diff
@@ -0,0 +1,186 @@
...
+import org.bukkit.event.player.PlayerUseUnknownEntityEvent;
+import org.bukkit.util.Vector;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+class PlayerBoundingBox implements Listener {
+
+    public static double maxDistance = 10;
```
twizmwazin: constant?

Pablete1234: fixed

twizmwazin:
merged

PR #996, opened 2016-04-09
Add a global XML to cardinal

Pablete1234:
Item mods module PR will add into the global xml, to add the axes changes that ocn made


src/main/java/in/twizmwaz/cardinal/util/DomUtil.java
```diff
@@ -14,9 +16,23 @@
 
 public class DomUtil {
 
+    private static Document global;
+
+    static {
+        try {
+            byte[] data = IOUtils.toByteArray(Cardinal.getInstance().getResource("global.xml"));
+            File globalFile = new File(Cardinal.getInstance().getConfig().getString("repo") + "/global.xml");
+            org.apache.commons.io.FileUtils.writeByteArrayToFile(globalFile, data);
```
Pablete1234: Is this the best way to go from resource / InputStream to file?

twizmwazin: No, I believe you can use plugin.getResoure("global.xml") and it will return an input stream. Also why was this not imported?

Pablete1234: plugin.getResource returns an input stream, but this code puts it into a file, is it the best way?

twizmwazin: FileInputStream?

Pablete1234: do i cast it?

twizmwazin:
merged

PR #1007, opened 2016-04-23
Prevent Title respawn Armor Stand showing health in GUI

Pablete1234:
It will make it so that the hearts of the entity you are ridding don't show up
It also fixes you being able to open crafting tables while you are dead
It fixes that when you switch teams, inventory doesn't update properly sometimes

Opinions?

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawn.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawn.java
@@ -1,6 +1,7 @@
 package in.twizmwaz.cardinal.module.modules.titleRespawn;
 
 import com.google.common.base.Optional;
+import com.google.common.collect.Lists;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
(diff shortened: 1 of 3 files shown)


Zenzoh:
Good job.

src/main/java/in/twizmwaz/cardinal/module/modules/titleRespawn/TitleRespawn.java
```diff
@@ -280,8 +283,24 @@ private void sendArmorStandPacket(Player player) {
...
+        // Create a packet to send attributes for the Armor Stand (with no modifiers)
+        PacketPlayOutUpdateAttributes attributePacket = new PacketPlayOutUpdateAttributes(Integer.MAX_VALUE, new ArrayList<AttributeInstance>());
+
+        try {
+            Field field = attributePacket.getClass().getDeclaredField("b");
+            field.setAccessible(true);
+
+            // Add a max health attribute with 0, so that health doesn't display
+            ((List<PacketPlayOutUpdateAttributes.AttributeSnapshot>) field.get(attributePacket)).add(
+                    attributePacket.new AttributeSnapshot("generic.maxHealth", 0D, Lists.<AttributeModifier>newArrayList()));
```
twizmwazin: You shouldn't need to use diamond types here. Is there a warning otherwise?

Pablete1234: Yes there is, it needs to be a Collection of attribute modifiers

Pablete1234:
merged

PR #1009, opened 2016-04-26
Add crafting module, fixes #1008

### LINKED ISSUE #1008: Crafting Recipes is not implemented yet
https://docs.oc.tc/modules/crafting

Pablete1234:
Everything should be done

Only needs a bit of testing

### DIFF
--- a/src/main/java/in/twizmwaz/cardinal/module/modules/craftingModule/recipes/ShapedRecipe.java
+++ b/src/main/java/in/twizmwaz/cardinal/module/modules/craftingModule/recipes/ShapedRecipe.java
@@ -1,0 +1,20 @@
+package in.twizmwaz.cardinal.module.modules.craftingModule.recipes;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.material.MaterialData;
+
+import java.util.Set;
+
+public class ShapedRecipe extends AbstractRecipe {
+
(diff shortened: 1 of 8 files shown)


Zenzoh:
Hey friend very very very good job with this, the truth is that you are a very competitive developer in the ambit of the competitivity on the Overcast Network competitive scrimmage plugins.

src/main/java/in/twizmwaz/cardinal/module/modules/craftingModule/recipes/ShapedRecipe.java
```diff
@@ -0,0 +1,27 @@
...
+        recipe = new org.bukkit.inventory.ShapedRecipe(result);
+        recipe.shape(shape);
+        for (Pair<Character, MaterialData> ingredient : ingredients) {
+            recipe.setIngredient(ingredient.getKey(), ingredient.getValue());
+        }
+    }
+
+    @Override
+    public void register() {
+        Bukkit.getServer().addRecipe(recipe);
```
iEli2tyree011: Since all recipes share the same functionality for the register method, you should create an AbstractRecipe class that you can extend.

Pablete1234: i should use bukkit recipes directly, but i didnt realize they were an interface before i made my own

Pablete1234:
Needs testing in all maps that use this if possible, i think it's ready (disabled recipes too)

Pablete1234:
merged

Issue #778, opened 2015-07-20
/rank info isn't implemented


TiTo-418:

twizmwazin:
What do we want this command to do? Just display the permissions?

TiTo-418:
The permissions and the ppl who have it?

twizmwazin:
Should it show online players or all players who have the rank?

iEli2tyree011:
@twizmwazin I think online players is fine, all players would get a little spammy. It should also show if the rank is a staff rank and if it's a default rank.

Pablete1234:
@iEli2tyree011 how about adding a ```/rank list player``` that displays the list of ranks the player has?

PR #1014, opened 2016-05-10
Implement `rank info` and update `rank list`.

iPGz:
Fix for the issue: #778 contains suggested features.

- Add rank information command (with flag '-i' for more info).
- Add player argument to `rank list` to display a players rank(s).

Video: https://streamable.com/cxgx


twizmwazin:
What exactly does `i` stand for, referring to the flag? Does it mean info? If it is showing further information, perhaps an `a` flag would be more appropriate.

iPGz:
Yeah I thought `i` for info would be best but `a` for all makes more sense, updated.

Pablete1234:
merged

PR #1023, opened 2016-06-06
Worldborder module (rebased)

Pablete1234:
Rebased world border module, to work in 1.9, tested in only 1 xml (without region files) and worked as expected


iPGz:
I remember having issues with this the last time I checked it out (a few months ago).
If it works as docs describes then it shouldn't be an issue.

Pablete1234:
@iPGz the problem is that docs aren't always too specific, but it should work as the docs say

PR #1025, opened 2016-06-08
Update to 1.10 & Add auto updater

Pablete1234:
Adds a "/cardinal -u" command that will try to update cardinal if a newer build is aviable. This will download lastest cardinal version, and you will only need to restart the server for the update to be complete

I want to add a setting that will automatically download it when you cycle if there is a new version, and just tells anyone with the cardinal.update permission that they should restart the server for the change to take effect


twizmwazin:
I would like to see this broken down into two commits, even if you want them merged together.

src/main/java/in/twizmwaz/cardinal/UpdateHandler.java
```diff
@@ -1,25 +1,46 @@
...
-    private final String localGitRevision;
     private boolean update;
 
     public UpdateHandler() {
         handler = this;
-        this.localGitRevision = GameHandler.getGameHandler().getPlugin().getDescription().getVersion().substring(GameHandler.getGameHandler().getPlugin().getDescription().getVersion().length() - 7, GameHandler.getGameHandler().getPlugin().getDescription().getVersion().length());
+        Bukkit.getScheduler().runTaskTimerAsynchronously(Cardinal.getInstance(), new Runnable() {
+            @Override
+            public void run() {
+                if (Cardinal.getInstance().getConfig().getBoolean("auto-update-cardinal")
```
twizmwazin: -cardinal really isn't necessary

Pablete1234:
Merged

PR #1028, opened 2016-06-10
Update Axes cool-down and damage.

iPGz:
[Updated](https://oc.tc/forums/posts/575ad03d5f35b97bd1000009) to match [global.xml](https://maps.oc.tc/Include/global.xml).


Pablete1234:
merged

PR #1029, opened 2016-06-11
Add team register

Pablete1234:
The old team register was completly broken by the name to UUID change, this one fixes therse issues.

PS: this is still by far not a perfect module, and i would like to rewrite parts of it

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,11 @@
             <artifactId>org.eclipse.jgit</artifactId>
             <version>4.5.0.201609210915-r</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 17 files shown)


twizmwazin:
Just like I said in the old PR, this isn't something I really want in Cardinal, because it directly ties it to OCN.

Pablete1234:
@twizmwazin I think there should be a way to abstract the team authenticator, where you give it a url, and a regex to find the player, if you want it to be searching in ocn, it would be "https://oc.tc/teams/%team%"  and search for the regex "avatar.oc.tc/%uuid%"

PR #1030, opened 2016-06-12
Fix developer list

Zenzoh:


RickAstIey:
vocuh good defender

Pablete1234:
This won't compile, when you fix the issue you can ammend the commit

src/main/java/in/twizmwaz/cardinal/module/modules/permissions/PermissionModule.java
```diff
@@ -39,7 +39,6 @@
             UUID.fromString("33a703d0-3237-4337-9ddd-3dbf33b3d8a6") /* iEli2tyree011 */,
             UUID.fromString("208c84af-790a-41da-bf7e-eb184f17bdf8") /* Elly */,
             UUID.fromString("260004f0-996b-4539-ba21-df4ee6336b63") /* Elliott_ */,
-            UUID.fromString("e82603df-417d-4a95-bd1f-1720648be0b4") /* Pablete1234 */);
```
Pablete1234: i don't think [this](https://github.com/twizmwazin/CardinalPGM/commits/master) aproves the change

Zenzoh:
Beaner

src/main/java/in/twizmwaz/cardinal/module/modules/permissions/PermissionModule.java
TiTo-418: approves*

PR #1041, opened 2016-06-17
Properly implement cause filters

Pablete1234:
Makes data station work as expected, not sure if all of the causes work the same as ocn.
It also adds events to the filter evaluation in some places it didn't, and tnt explosions also filter the player that placed the tnt,


Pablete1234:
merged

PR #1042, opened 2016-06-17
Fix mirrored and translated regions

Pablete1234:
mirrored regions in boom box already were working, but dataStation ones weren't for some reason, so i decided to rewrite how mirrored regions are handled, now they work perfectly fine both in boombox and in datastation


Pablete1234:
merged

PR #1046, opened 2016-07-11
Add team option to proximity

Pablete1234:
This is good for maps like battle ecliptica, in the scoreboard you can't see your proximity, but you can /proximity [your team] and you will always have permission to do that.

Also, the "hasObjectives" thing was redone, it now shows every team if the team has any shown objective, so maps like Larnia, where only one team has an objective, it doesn't display all teams (defenders being empty, and attackers having one flag).


Pablete1234:
merged

PR #1050, opened 2016-09-16
Add nbt interact method

Pablete1234:
It adds a method to get a cardinal data tag, where we can add/read stuff from an item, also, item mods is updated to use it, and some missing attributes and elements have been added in item mods (can place on, can destroy and show flags)


Pablete1234:
merged

PR #1052, opened 2016-09-26
Rewrite tab list

Pablete1234:
Basically moved lots of code arround, organized the whole thing, and made the base better, and less prune to errors


Pablete1234:
merged

PR #1057, opened 2016-10-09
Fix proximity on completed objectives

Doguillo:
Exclude completed objectives when calculating the winning team (don't count them twice).


Pablete1234:
Merged

PR #1058, opened 2016-10-12
Add multiple repository support, and add git repositories

Pablete1234:
It replaces the old "repo" config with a "repos" config, that contains a list of repositories, such list can have a list of folders, or git repositories:

```
repos:
- maps
- git:https://github.com/User/Repo
- git:https://login@pass/User/APrivateRepo
```

This example would have 2 repositories, one would be the maps folder, the other one would be a git repository from the user "User" named "Repo" hosted on github

Git repositories will only update when you run /newmaps, so if any git repository pushed a map change, you need to reload maps in order to be able to use such modification

Suggest other repository formats and i'll see what i can do to implement them


CoWinkKeyDinkInc:
Thanks for adding this awesome feature!  Will this clone the repo or would it just download the maps?  Does using this feature make it work with Git clients/the terminal?

Pablete1234:
@CoWinkKeyDinkInc This will git clone the map repository if not already downloaded, and will git reset to latest commit every time you run /newmaps or on server restart, without the need of re-cloning everything, regarding forks, i'm not sure if github has ways of referencing branches inside the url

CoWinkKeyDinkInc:
Alright, thanks!  This thing should work with forks, but yeah, I think implementing branches might be tricky, but it isn't convenient for the user anyway, there's multiple repositories so branches would be pointless.  If it resets to the latest commit every-time, than it looks like this won't be handy for bug fixing maps, but I can just manually clone the fork.  I'm so excited to see this finally implemented!!

Pablete1234:
@CoWinkKeyDinkInc for map testing you should most likely be working in local files, inside the local repo, that will update by just recycling the map, instead of needing to use newmaps, when it's done, you can push it

Pablete1234:
Because maps can sometimes be duplicated, or it may be hard to manage maps with similar names, i added map id's, the id's reference a map in an static moment, and will change if you run /newmaps.
To see id's, `map` & `maps` commands now have a `-l` (long) flag.
The long flag will also show repository and folder in the `map` command.
You can use the id's in `map`, `cycle` and `setnext` commands, with the `-m` (map) flag.
Ex: `cycle 10 -m 15` will cycle in 10 seconds to the map with id 15

Prodigy4532:
Work with privated repositories?

Pablete1234:
@Prodigy4532 they work on github if you do:
git:https://user:password@github.com/uset/repo
But don't use that for public servers yet as /map -l will show the source repo, including user and pass

Prodigy4532:
Thanks

Pablete1234:
@Prodigy4532 finally coded private git repos, using:
`git:https://user:password@github.com/user/repo`
will work fine, and should work fine with other git repository services (like gitlab), even though i didn't test it, running a /map -l will show the repository path without user or pass info (`https://github.com/user/repo`)

twizmwazin:
It would probably be best if this is reorganized, commit wise. Generally if the word "and" appears in a commit message that's a red flag that it should be split.

Pablete1234:
@twizmwazin If the title is the issue i can fix that by changing it to "Refactor everything related to repositories"

twizmwazin:
Did you write this in such a way that multiple repository support demands the presence of git repositories? Small changes are important, it makes it easier to read through, both for review and accountability.

twizmwazin:
merged

PR #1063, opened 2016-11-05
Refactor config

Pablete1234:
Make config be loaded from and to a config util.

Changes the old:
`Cardinal.getInstance().getConfig().getBoolean("aSettingName")`
to:
`Config.aSettingName`

Also, we no longer need the config.yml file, as all the setting defaults are obtained from the Config util fields


Pablete1234:
merged

PR #1064, opened 2016-11-10
Rework countdowns, add defaults to config

Pablete1234:
Config now has: `startDefault` and `cycleDefault` that will be the defaults for when someone runs cycle, recycle, start, without a time, or when all teams ready,


twizmwazin:
merged

